PR #9, opened 2018-02-05
add test for tutorial value

gidden:
This shows an example of how to directly test tutorial values.


danielhuppmann:
very nice, thanks @gidden !

PR #10, opened 2018-02-09
replace GAMS-InputDir by `subprocess(cwd=...)`

danielhuppmann:
GAMS looks for opt-files and writes output files to the working directory irrespective of the option InputDir. This PR replaces the InputDir argument in the `solve()` function calling the subprocess by the cwd argument of the `check_call()`.


danielhuppmann:
merged myself after another round of testing...

PR #11, opened 2018-02-14
add filters when retrieving timeseries data from database

danielhuppmann:
This PR adds the option to apply filters when retrieving timeseries data for a Scenario or TimeSeries object.

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -345,11 +345,11 @@ class TimeSeries(object):
 
         Parameters
         ----------
-        IAMC : boolean, default True
+        iamc : boolean, default True
             returns a pandas.DataFrame either
             - 'IAMC-style' format (cols: region, variable unit, <years>)
             - in tabular form (cols: region, variable, unit, year)
-        region : list of strings
+        regions : list of strings
             filter by regions
         variables : list of strings
             filter by variables
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -20,6 +20,12 @@ import ixmp.model_settings as model_settings
 
 local_path = os.path.expanduser(os.path.join('~', '.local', 'ixmp'))
 
+# %% common definitions
+
+iamc_idx_cols = ['model', 'scenario', 'region', 'variable', 'unit']
+
+
+# %% Java Virtual Machine start-up
 
 def start_jvm():
     if jpype.isJVMStarted():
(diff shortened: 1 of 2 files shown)


ixmp/core.py
```diff
@@ -333,41 +339,63 @@ def add_timeseries(self, df, meta=False):
                 self._jobj.addTimeseries(df.region[i], df.variable[i], time,
                                          jData, df.unit[i], meta)
 
-    def timeseries(self, IAMC=False, node=None, variable=None):
-        """retrieve data from this TimeSeries instance as a pandas dataframe
+    def timeseries(self, iamc=False, regions=None, variables=None, units=None,
```
OFR-IIASA: Just some small consistency issues:
in the function iamc is in small letters while it is written in capital in the description below. In the function regions; in the description below regions.

tests/test_core.py
```diff
@@ -166,23 +167,44 @@ def test_years_active_extend(test_mp):
...
 def test_new_timeseries(test_mp):
     scen = test_mp.TimeSeries(*test_args, version='new', annotation='testing')
-    df = {'region': ['World'], 'variable': ['Testing'], 'unit': ['???'],
-          'year': [2020], 'value': [23.6]}
+    df = {'year': [2010, 2020], 'value': [23.5, 23.6]}
     df = pd.DataFrame.from_dict(df)
+    df['region'] = 'World'
+    df['variable'] = 'Testing'
+    df['unit'] = '???'
     scen.add_timeseries(df)
```
OFR-IIASA: I have used the same formulation as above df['xyz'] = 'abc' and always receive a warning from python 3.6 when doing this. An alternative way of writing this which seems to be safer according to the link below is df.loc[:,'XYZ'] = 'abc'
see below
"A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  df['unit'] = 'GWa'"

danielhuppmann: I know this error message happens in some instances, but I don't get it here... I guess it's not a problem here because this is a new dataframe (as opposed to an dataframe cut from an existing one)?

danielhuppmann:
this PR depends on the `ixmp.jar` from #17

PR #18, opened 2018-02-20
alllow alternative header cases when adding timeseries

danielhuppmann:
This PR checks that all columns required for the IAMC template are present when adding timeseries data to an `ixmp.TimeSeries()` object.

The function now accepts "Title" and "UPPER" header styles, and accepts 'node' instead of 'region'.

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -22,7 +22,7 @@ import ixmp.model_settings as model_settings
 
 local_path = os.path.expanduser(os.path.join('~', '.local', 'ixmp'))
 
-# %% common definitions
+# %% default settings for column headers
 
 iamc_idx_cols = ['model', 'scenario', 'region', 'variable', 'unit']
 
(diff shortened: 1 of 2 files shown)


gidden:
Lgtm, looks like you caught some tests though =)

danielhuppmann:
`model` and `scenario` are obvious because the data is added to an `ixmp.TimeSeries` or `ixmp.Scenario` instance

danielhuppmann:
@gidden, fixed the test...

gidden:
Final request - one short test which raises your value error?

danielhuppmann:
Good call, @gidden - rebasing and adding the test failed because the previous PR from @ClaraLuisa  caused issues when you had `psutils` installed, but call the `Platform()` with `jvmargs` - because it tried to set them to `-Xmx3.98723734G`, which Java didn't like...

Fixed it by casting the auto-jvmsize to `int`.

Should be good to go, now!

gidden:
Awesome!

PR #26, opened 2018-03-13
Java virtual machine Xmx can now be user specified

ClaraLuisa:
If not specified by the user, the half available memorry is allocated as upper limit to the vitual machine intead of static 4G as before.

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -29,7 +29,7 @@ iamc_idx_cols = ['model', 'scenario', 'region', 'variable', 'unit']
 
 # %% Java Virtual Machine start-up
 
-def start_jvm(jvmargs):
+def start_jvm(jvmargs=None):
     if jpype.isJVMStarted():
         return
 
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -33,15 +33,13 @@ def start_jvm(jvmargs=None):
     if jpype.isJVMStarted():
         return
 
-    try:
-        import psutil
-        if jvmargs is None:
+    if jvmargs is None:
+        try:
+            import psutil
             jvmsize = psutil.virtual_memory().available / 10**9 / 2        
             jvmargs = "-Xmx{}G".format(jvmsize)        
-    except ImportError:
-        if jvmargs is None:
+        except ImportError:
             jvmargs = "-Xmx4G"              
-        pass    
 
     # must add dir and jarfile to support finding ixmp.properties
     module_root = os.path.dirname(__file__)
@@ -81,7 +79,6 @@ class Platform(object):
         the type of the local database (e.g., 'HSQLDB')
         if no 'dbprops' is specified, the local database is
         created/accessed at '~/.local/ixmp/localdb/default'
-        
     jvmargs : string
         the allocated max heap space for the java virtual machine
         eg.: "-Xmx4G"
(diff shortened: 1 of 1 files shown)


danielhuppmann:
thanks @ClaraLuisa for this nice improvement!

We could make `psutils` optional, setting `-Xmx1G` if the package can't be loaded, and detect it automatically if `psutils` is available.

Also, I'd suggest to pass over the entire instruction as an argument, not just the number. So have `jvmargs='-Xmx1G'`, as default, and require users to pass over `jvmargs='-Xmx12G'` if they want to work with really big data stuff...

ClaraLuisa:
@danielhuppmann, thanks for that idea. I made psutil optional now as recommended by you. However I set the default to 4 instead of 1G as I running it wasn't very stable for me otherwise. Is this acceptable for you?

ClaraLuisa:
...and do know why the CI test keeps failing. The code seems to be working well on my PC.

ixmp/core.py
```diff
@@ -27,18 +29,28 @@
 
 # %% Java Virtual Machine start-up
 
-def start_jvm():
+def start_jvm(jvmargs=None):
     if jpype.isJVMStarted():
         return
 
+    try:
```
gidden: only a minor comment, can you make this a bit more concise, like:

```
if jvmargs is None:
    try:
        jvmargs = ...
    except:
        jvmargs = ...
```

gidden: And `pass` at the end is not needed

ixmp/core.py
```diff
@@ -69,10 +81,14 @@ class Platform(object):
         the type of the local database (e.g., 'HSQLDB')
         if no 'dbprops' is specified, the local database is
         created/accessed at '~/.local/ixmp/localdb/default'
+
```
gidden: no space needed in docs

ClaraLuisa:
Thnaks @gidden for the hints!

ixmp/core.py
```diff
@@ -69,10 +79,13 @@ class Platform(object):
         the type of the local database (e.g., 'HSQLDB')
         if no 'dbprops' is specified, the local database is
         created/accessed at '~/.local/ixmp/localdb/default'
+    jvmargs : string
+        the allocated max heap space for the java virtual machine
+        eg.: "-Xmx4G"
```
gidden: Final note, can you add a link to the Java CLI docs: https://docs.oracle.com/javase/7/docs/technotes/tools/windows/java.html so users can look up themselves other options? Then that's it for me!

gidden:
Final comment by me inline, should be good to go by my eye after!

ClaraLuisa:
@gidden I am not sure if this is what you expected with "comment by me inline". Let me know if anything else should be changed.

gidden:
Yup that's it! lgtm, @danielhuppmann want to pull?

danielhuppmann:
very nice, thanks @ClaraLuisa and @gidden

ClaraLuisa:
Hi @danielhuppmann and @gidden did this merge also affect the submodule ixmp in the message_ix repository? Bc. even though i updated the submodule the new code does not appear in the core.py file. What am I doing wrong?

danielhuppmann:
For future reference - updates in `ixmp` need to be explicitly included in `message_ix` by fast-forwarding the submodule and committing the change to `message_ix` - see https://github.com/iiasa/message_ix/pull/5 for the PR pulling the change here to `message_ix`.

PR #27, opened 2018-04-05
added 2 tests to test timeseries variable name length

OFR-IIASA:
When adding a timeseries with a variable which has more than 64 characters, an error message is displayed. The tests below replicate this issue

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -176,6 +176,34 @@ def test_new_timeseries(test_mp):
     scen.commit('importing a testing timeseries')
 
 
+def test_new_timeseries_long_name64(test_mp):
+    scen = test_mp.Scenario(*msg_multiyear_args)
+    scen = scen.clone(keep_sol=False)
+    scen.check_out(timeseries_only=True)
+    df = pd.DataFrame({
+    'region': ['India', ],
+    'variable': ['Emissions|CO2|Energy|Demand|Transportation|Aviation|Domestic|Fre',],
+    'unit': [ 'Mt CO2/yr', ],
+    '2012': [ 0.257009, ]
+    })
+    scen.add_timeseries(df)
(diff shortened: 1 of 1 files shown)


OFR-IIASA:
Currently, there is an issue with variable names longer than 64 characters being uploaded as timeseries. This is an issue for India and needs to be resolved. I am not sure what the cause is and how this can be resolved.

danielhuppmann:
I had a look at this on the Java-source side and didn't find the cause. Will look at this later...

gidden:
Perhaps @peterkolp or @zikolach can provide insight here?

danielhuppmann:
hi @zikolach, can you check whether this problem still occurs after moving to the DAO implementation?

peterkolp:
(a bit guessing) Does the variable name end's up in IX_KEY.KEY? 
which is limited to 40 bytes .,.

danielhuppmann:
good call @peterkolp , but no... That's only for set elements.

peterkolp:
I see. 
AND we do have longer than 64 char timeseries names in database.
In SENSES_TEST schema the actual max(length(keystring)) from iamc_key is 80. 
In ENE_IXMP schema it's 82.

So RDBMS and java core seem to be able to handle >64 char strings (up to 256).

danielhuppmann:
Correct, @peterkolp , and the setup of the HSQLB also does set 265 as string-size of the `IAMC_KEY` table, so that shouldn't be a problem either...

zikolach:
Table `change_log` column `item` is 64 characters long (` ITEM VARCHAR(64)  not null`)

zikolach:
@danielhuppmann yes, it will take care (see e9978d5ff28c432ec93ff4d5b5b8513ab1503614)

gidden:
@OFR-IIASA can you rebase this on most recent master?

PR #31, opened 2018-05-14
Clean setup

zikolach:


README.md
```diff
@@ -1,5 +1,16 @@
 # The ix modeling platform (ixmp)
 
+## TLDR
+
+```bash
+# setup python2/3 virtual environment
```
gidden: Perhaps add a note that this is optional? It is also not exactly a tl;dr because users will have to install all the other non-trivial dependencies. 

@danielhuppmann should comment on how/where he thinks this should be placed.

danielhuppmann:
I think we don't want this as the first item you see when you look at the repsoitory. I'd add that as an option or additional entry in the dependency installation that you can use the virtual environment...

requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+numpy
+pandas
+pytest>=3.0.6
+JPype1>=0.6.2
+sphinx
+sphinxcontrib-bibtex
+sphinxcontrib-fulltoc
+numpydoc
+cloud_sptheme
+nbformat
```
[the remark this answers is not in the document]
zikolach: added it

.gitignore
```diff
@@ -40,10 +40,17 @@ config/*.properties
...
 rixmp/source/inst/docum/*.html
 rixmp/source/.Rhistory
 !rixmp/source/*.Rproj
 .Rproj.user
+
+# IDEA
+.idea/
+
+# Python virtualenv
+.env*/
```
danielhuppmann: missing eol

danielhuppmann:
@gidden, how does this interact with your tackling a better install procedure?

gidden:
The instructions here still provide source builds. We will probably want to revamp these instructions in general with the updated conda installs.

gidden:
@zikolach do you want to add the `.dylib`s you need to this PR as well?

zikolach:
@gidden yes, sorry I thought I pushed it already. Now it is in PR

gidden:
Hey @zikolach, we've had a lot of thrashing on the master branch recently before the YSSPers get here. Would you mind opening a new PR with changes you still think as relevant?

PR #34, opened 2018-05-28
retixmp (reticulate) + 2 rixmp installations tips

adrivinca:
The 'retixmp' package and a script to build it are added in the main folder.
the package allows to pass every dataframe or array to ixmp, using the function `adapt_to_ret()`, without paying much attention to the type of unit. 
e.g.

`scenario$add_par('inv_cost', adapt_to_ret(df)`

Transport tutorials are added, but they are exactly equal to the R or python tutorial. no functions specific of python-ixmp are used.

Then, I also added to the readme two tips for avoiding issued when installing rixmp


gidden:
Should the the .source and .html files be added here?

adrivinca:
I cleaned a bit the folder and left only those files needed in case we want retixmp to substitute rixmp, or even just add to in the one-click installer

gidden:
ok, lgtm. @adrivinca can you sign the CLA then we can merge?

adrivinca:
i signed, if it is ok. we can merge. So that I can add the austria-reticulate tutorial in message_ix_dev once the ixmp sub-module is updated. thanks

gidden:
Ok, lgtm!

PR #44, opened 2018-06-01
add a hook to also build docs on ci

gidden:


danielhuppmann:
lgtm, thanks @gidden !

PR #47, opened 2018-06-05
added 2 tests to test timeseries variable name length

gidden:
rebased #27


gidden:
huzzah! thanks @OFR-IIASA @zikolach @peterkolp and @danielhuppmann

PR #51, opened 2018-06-25
add lru cache to certain scenario calls

gidden:
@OFR-IIASA this should do the trick

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -8,6 +8,12 @@ import pandas as pd
 from jpype import JPackage as java
 from subprocess import check_call
 
+try:
+    from functools import lru_cache  # py3
+except ImportError:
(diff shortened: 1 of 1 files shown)


danielhuppmann:
we had this discussion before - in short, I don't see why we need a duplication of the pycache feature - if you call `scenario.element(name)` without filters, the entire data (as pd.DataFrame) is stored in the python cache. Calling that function again with filters just applies the selection to that dataframe, and I guess pandas is quite efficient.

Also, pycache implements scenario changes - so if data of a scenario is changed (check-out, edit, commit), the pycache is cleared. Switching to lru_cache might circumvent version control.

gidden:
Yes, I forgot that we were still using the pycache I wrote. Should probably
discuss briefly with @OFR-IIASA on his profiling efforts here.

PR #52, opened 2018-06-25
add function to completely load scenario in memory

gidden:


gidden:
for comparison with #51

gidden:
should be ready for review now

PR #54, opened 2018-06-25
Dynamic DB

zikolach:
PR is containing 2 features:
- dynamic database compatibility
- exposing metadata API to Scenario

Changes:
- updated ixmp.jar
- added tests for cloning scenario
- added metadata api methods to Scenario (iiasa/ixmp_source/#79)
- added name parameter of get_meta
- added metadata set/get test


gidden:
@zikolach can you suggest which test or tests should be implemented here to test the behavior?

zikolach:
@gidden it may be simple scenario copy test like:

1. Create first Platform instance (providing first properties file path for first db)
2. Load existing scenario
3. Create another Platform instance (providing different properties file path with connection details for second db)
3. Call clone on scenario instance (from 2) providing platform instance (from 3)
4. Ensure that cloned scenario (returned by from clone method) contains all the same as original one

In best case target database should contain different identifiers in ix_unit table to check performed mapping.

gidden:
tests added in @zikolach's fork PR. @zikolach can you please remove the `force_migration` lines and confirm all still works as expected?

gidden:
To note, we will also want to add this to our (currently non-existent) nightly tests to confirm behavior between oracle and localdbs.

gidden:
It looks like the error here is with implicit kwargs. @zikolach can you move the `platform` argument in `clone()` to the end of the kwarg list and update the docstring? All should pass then.

gidden:
can you rebase on master @zikolach ? That will fix stickler-ci

gidden:
Can you rebase this on master please @zikolach ?

ixmp/core.py
```diff
@@ -1017,9 +1018,21 @@ def years_active(self, node, tec, yr_vtg):
         """
         return to_pylist(self._jobj.getTecActYrs(node, tec, str(yr_vtg)))
 
+    def get_meta(self, name=None):
+        return {x.getKey(): unwrap(x.getValue()) for x in
```
gidden: should probably just use a lambda here:

```
unwrap = lambda x: x if type(value).__name__ != 'java.math.BigDecimal' else x.doubleValue()
```

is it possible to use `isinstance()` instead of `type().__name__`?

gidden: also easier to understand by adding var:

```
meta = np.array(self._jobj.getMeta().entrySet().toArray()[:])
```

tests/test_core.py
```diff
@@ -227,3 +226,21 @@ def test_timeseries_edit(test_mp_props):
...
+def test_add_meta(test_mp):
+    scen = test_mp.Scenario(*can_args, version=1)
+    scen.set_meta('test_string', 'test12345')
+    scen.set_meta('test_number', 123.456)
+    scen.set_meta("test_number_negative", -123.456)
+    scen.set_meta('test_int', 12345)
+    scen.set_meta('test_bool', True)
+    scen.set_meta('test_bool_false', False)
+    meta = scen.get_meta()
+    npt.assert_equal(meta, dict([
```
gidden: we use the testing pattern `obs` and `exp`, so this can be

```
obs = scen.get_meta()
exp = {
  'key': value,
  ...
}
```

tests/testdb/test.properties_template
```diff
@@ -5,6 +5,8 @@ jdbc.url.1 = jdbc:hsqldb:file:{here}/ixmptest
 jdbc.user.1 = ixmp
 jdbc.pwd.1 = ixmp
 
+force_migration = false
```
gidden: please remove

gidden:
A few suggestions from my end, but @danielhuppmann should take a look as well as I have commits in the PR..

ixmp/core.py
zikolach: I didn't find other way to check Java type for JPype

gidden:
hey @zikolach, I also saw and fixed the linter issues with fixtures in #52. No need to do anything with that. @danielhuppmann this should now be ready for your review.

ixmp/core.py
```diff
@@ -860,11 +858,13 @@ def equ(self, name, filters=None, **kwargs):
         return self.element('equ', name, filters, **kwargs)
 
     def clone(self, model=None, scen=None, annotation=None, keep_sol=True,
-              first_model_year=None):
+              first_model_year=None, platform=None):
         """clone the current scenario and return the new scenario
 
         Parameters
         ----------
+        platform : ixmp.Platform
```
gidden: @zikolach can you move this docstring down to the bottom to match the function api?

ixmp/core.py
```diff
@@ -1017,10 +1018,23 @@ def years_active(self, node, tec, yr_vtg):
         """
         return to_pylist(self._jobj.getTecActYrs(node, tec, str(yr_vtg)))
 
+    def get_meta(self, name=None):
```
gidden: needs a docstring

ixmp/core.py
```diff
@@ -1017,10 +1018,23 @@ def years_active(self, node, tec, yr_vtg):
...
+        meta = np.array(self._jobj.getMeta().entrySet().toArray()[:])
+        return {x.getKey(): unwrap(x.getValue()) for x in meta}
+
+    def set_meta(self, name, value):
```
gidden: needs a docstring

ixmp/core.py
```diff
@@ -1017,10 +1018,23 @@ def years_active(self, node, tec, yr_vtg):
...
+        self._jobj.setMeta(name, value)
 
 # %% auxiliary functions for class Scenario
 
 
+def unwrap(value):
```
gidden: prefer defining this in the single function it is needed in, it's such a special case, I don't think it will be needed elsewhere, and if it is, it should probably be made more generic (e.g., jtype_to_pytype)

gidden:
apologies, a few more changes suggested.

gidden:
Hi @zikolach, can you `git commit --amend` and `git push -f`? That should re-instigate the checks.

gidden:
closing in favor of #60

gidden:
@zikolach, let's please build new PRs on top of this as #61 does

PR #73, opened 2018-07-27
Clean error message for `check_out()` and refactor to `..._solution`

### LINKED ISSUE #71: Remove solution error message
The exception error for removing solution (for example, when trying to check out a scenario that contains a solution) suggests using the command 'removeSolution()'. However, the actual command doing this task now is 'remove_sol()'. It seems the text in the error message needs to be updated.

danielhuppmann:
This PR migrates the error message when checking out a Scenario with a solution from Java to Python, to get the correct function names: `remove_solution()` instead of the underlying Java function `removeSol()`.

In addition, the PR adds a function `has_solution()` and renames `remove_sol()` to `remove_solution()`. The kwarg of `clone()` is changed from `keep_sol` to `keep_solution`, but `keep_sol` continues to work (with a deprecation warning).

Closes #71

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -859,8 +859,8 @@ class Scenario(TimeSeries):
         """
         return self.element('equ', name, filters, **kwargs)
 
-    def clone(self, model=None, scen=None, annotation=None, keep_sol=True,
-              first_model_year=None):
+    def clone(self, model=None, scen=None, annotation=None, keep_solution=True,
+              first_model_year=None, **kwargs):
         """clone the current scenario and return the new scenario
 
         Parameters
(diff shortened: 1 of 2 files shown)


danielhuppmann:
CircleCI passed with tests using deprecated `clone(..., keep_sol=False)` formulation. Now updated the tests formulation to `keep_solution=False` to avoid `deprecated`-warning when running continuous integration.

ixmp/core.py
```diff
@@ -917,23 +929,31 @@ def read_sol_from_gdx(self, path, filename, comment=None,
...
+        """check whether the Scenario has been solved and a solution (variables
+        and equations) exists in the database"""
+        return self._jobj.hasSolution()
 
-    def remove_sol(self):
+    def remove_solution(self):
         """delete the solution (variables and equations) from the sceanario"""
-        self.clear_cache()  # reset Python data cache
-        self._jobj.removeSolution()
+        if self.has_solution():
```
gidden:
```
if not self.has_solution():
   raise
...
```

gidden:
Ok, just going to merge, not worth changing at this point. Thanks @danielhuppmann!

gidden:
hey @danielhuppmann, I forgot to think about the release notes. probably worth adding a note that there is a new `has_solution()` function..

PR #79, opened 2018-07-31
adding warning for calling class ixmp.Scenario with `scheme=='MESSAGE'`

### LINKED ISSUE #77: add deprecation warning to `ixmp.Scenario(..., scheme='MESSAGE')`
Using the class `ixmp.Scenario` for MESSAGE-scheme scenarios is deprecated and should be switched to the MESSAGE-specific sub-class `message_ix.Scenario`.

danielhuppmann:

- [ ] Tests Added - **No additional tests required**
- [ ] Documentation Added - **No additional documentation required**

# Description

This PR raises a deprecated-warning whenever a (generic) `ixmp.Scenario` is called that is defined as a MESSAGE-scheme scenario. This warning also works when loading existing scenarios from a database.

closes #77

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- [#79](https://github.com/iiasa/ixmp/pull/79): Adding a deprecated-warning for `ixmp.Scenario` with `scheme=='MESSAGE'`
 - [#76](https://github.com/iiasa/ixmp/pull/76): Changing the API from `mp.Scenario(...)` to `ixmp.Scenario(mp, ...)`
 - [#73](https://github.com/iiasa/ixmp/pull/73): Adding a function `has_solution()`, rename kwargs to `..._solution`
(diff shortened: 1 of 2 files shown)


gidden:
@danielhuppmann, just had a thought... we actually use this constructor with the `scheme` argument in `message_ix`... so all our users will see this. We will probably have to wait to deprecate this until after we pull everything out of java..

danielhuppmann:
couldn't help myself, @gidden - not the most elegant solution, but it seems to work without too much fuss or circular dependencies...

corresponding PR in `message_ix`- https://github.com/iiasa/message_ix/pull/90

gidden:
ok, seems good to me. perhaps we should add a comment to remove on next release?

danielhuppmann:
I would have thought that prior to the next release, we just search for the word `deprecate`... or should this go to the release notes or another document explicitly?

gidden:
sounds good!

PR #81, opened 2018-08-22
features requiring new jar file

gidden:
nominally 

- multiple database connections
- fetching metadata directly from java object
- setting log levels

**NOTE** this still needs to most recent jar file to be added before tests should pass


gidden:
ok, @danielhuppmann I think this is all of the python features we need with the new jar file. To note, this will fail for the moment, as we actually need the jar file in this PR with all of the features in (e.g., `setLogLevel`)

gidden:
closing in favor of #88

PR #86, opened 2018-09-10
updating db migration for hsql and macro support

gidden:
as stated above


gidden:
merging to keep working on iiasa/message_ix#109

PR #88, opened 2018-09-13
multiple platform-(db)-connections, MESSAGE-scheme v1.1

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description

This PR introduces the option to connect to multiple databases (multiple instances of an `ixmp.Platform` ), with the option to clone across platforms. It also makes logging less noisy and adds an option to set the logger-level via Python.

Furthermore, it upgrades the MESSAGE-scheme default sets and parameters incorporating the changes made during the hackathon at IIASA in August 2018.


gidden:
I guess this still needs to fix the migration issue? should we remove the V1.6 or somehow rename it?

danielhuppmann:
It's probably easier to change the flyway-migration files in `ixmp_source`, since this has only touched some test databases for the scenario explorer. The flyway-checksum issue that makes CI fail is caused by platform-dependent line endings, still need to fix that...

gidden:
Ok, want to ping me when it's ready to review/bring in?

danielhuppmann:
please keep on hold until we discuss how to treat "drivers"...

gidden:
we're still waiting for the emissions stuff here right?

danielhuppmann:
yes, @gidden, still have to test the emissions implementation...

danielhuppmann:
@gidden, any ideas why the notebook execution tests would fail only on AppVeyor and only on Python 3...? All tests pass on my work PC, which would actually be closest to this CI configuration...

gidden:
No idea, trying to rebuild. Sometimes fishy things can happen.

gidden:
initial test in #93. basically, I pulled back to the last successful commit (fc12854) to see if it still works. Let's diagnose after that goes through.

danielhuppmann:
thanks @gidden for the assist! closes #93

gidden:
Ok, this seems fine to me. I believe this is our final PR for v0.2.0. Feel free to let me know if this is not the case @danielhuppmann

danielhuppmann:
I think so, @gidden! Nothing I can think right now of that still needs to be done urgently.

PR #94, opened 2018-11-13
Update jarfile to current ixmp_source master

zikolach:
- replace jarfile (compiled with jdk8)
- fix migrations in test DB (ixmptest.script)
- add start HSWLDB client script


gidden:
thanks @zikolach !!

PR #99, opened 2018-11-21
migrate to versioneer

gidden:
Previously, we had to define version numbers in multiple places, this uses the `versioneer` tool to do it automatically from git tags and commits.


gidden:
tracking the prior commit [here](https://ci.appveyor.com/project/danielhuppmann/ixmp/builds/20456222)

once it passes, will pull

PR #103, opened 2018-12-04
Add dependencies in setup.py

khaeru:
Parallel to iiasa/message_ix#139. This PR:

- Sets the "install_requires" and "extras_require" keyword arguments in `setup.py`. This ensures that the dependencies automatically installed with the package.
- Trims `requirements.txt` to remove old dependencies.
- Changes the versioneer style to `pep440-pre` from `git-describe`. See explanation [here](https://github.com/iiasa/message_ix/pull/139#issuecomment-444428968).

**PR checklist:**
- [x] Tests added
- [x] Documentation added

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ INSTALL_REQUIRES = [
 ]
 
 EXTRAS_REQUIRE = {
-    'tests': ['pytest>=3.0.6', 'nbformat', 'jupyter'],
+    'tests': ['pytest>=3.0.6', 'jupyter'],
     'docs': ['cloud_sptheme', 'numpydoc', 'sphinx', 'sphinxcontrib-bibtex',
              'sphinxcontrib-fulltoc'],
     'tutorial': ['jupyter'],
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,6 @@ from setuptools import setup
 
 
 INSTALL_REQUIRES = [
-    'cython',
     'JPype1>=0.6.2',
     'pandas',
 ]
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,8 @@ from setuptools import setup
 INSTALL_REQUIRES = [
     'JPype1>=0.6.2',
     'pandas',
+    'xlsxwriter',
+    'xlrd',
 ]
 
 EXTRAS_REQUIRE = {
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,24 @@
 #!/usr/bin/env python
 from __future__ import print_function
 
-import os
-import shutil
-import sys
 import glob
 import versioneer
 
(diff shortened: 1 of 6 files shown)


khaeru:
[Same question applies](https://github.com/iiasa/message_ix/pull/139#issuecomment-444207776) before this is ready to go: is `cython` a dependency here (where?), or of message_ix? In the latter case, delete it from the list.

khaeru:
@gidden [replied on the other PR](https://github.com/iiasa/message_ix/pull/139#issuecomment-444384937):
> Cython is probably no longer a dependency (used to be)
…so I will try removing it.

setup.py
```diff
@@ -40,6 +50,8 @@ def main():
                   'Oliver Fricko, Peter Kolp',
         "author_email": 'message_ix@iiasa.ac.at',
         "url": 'http://github.com/iiasa/message_ix',
+        "install_requires": INSTALL_REQUIRES,
```
zikolach: @khaeru I am not so familiar with setting up python packages. Could you explain a bit how does `install_requires` and `extra_require` work? Is it now not necessary to run `pip install -r requirements.txt` - so `python setup.py install` will do the magic?

khaeru: `pip install .` should be the only installation command. This will:
1. Download and install all the packages from `install_requires`, if the user doesn't already have them.
2. Implicitly call `setup.py install` to install the package in the current directory (".").

Calling (for instance) `pip install .[docs]` means "install the current package, and also the 'docs' extra features." So the dependencies specified in `extras_require['docs']` will also be downloaded and installed. Multiple extras can be specified: `pip install .[docs,tests]`.

**More details:** Using a requirements.txt is okay for CI tools, where one can give a script that does many things as part of the installation process.

But an end-user installing the package from PyPI will run `pip install ixmp`. In this case, pip will only understand dependencies given in setup.py.

I think this issue hasn't come up before now because Anaconda was preferred for ixmp/message_ix. Anaconda packaging has an entirely different format/workflow, so [the dependencies are specified in a different place](https://github.com/gidden/ixmp-feedstock/blob/master/recipe/meta.yaml#L24) and it didn't make a difference that they were missing from setup.py.

zikolach: Thanks for explanation. I added `nbformat` and `jupyter` as extra dependencies for tests to `setup.py` (see last commit) to make tests working. Should we update readme to replace `pip install -r requirements.txt` with `pip install .`?

khaeru: Just pushed a commit updating the README.

I think that `nbformat` is a dependency of `nbconvert` which is in turn a dependency of `jupyter`. So probably just `jupyter` would be sufficient; did you try that?

zikolach: Yes, just tried and it worked. Thanks!

PR #106, opened 2018-12-13
Add new jar for testing

zikolach:
I would like someone to test new jar file created from iiasa/ixmp_source#141.

It incorporates lots of changes, in particular version from PR uses connection pool to access DB instead of single connection (methods which were previously dependent on using single database connection and couldn't work in parallel in single process/app instance now changed to take single connection from the pool and give it back as soon as the work is done).

If possible I'd like someone to test performance of DB operations (e.g. on loading/saving large scenarios) to compare current release version and new one.


zikolach:
Closing in favor of #116

PR #115, opened 2019-01-23
Support iterating with user-supplied callbacks

khaeru:
This PR adds a feature to `Scenario.solve()` iteratively, with some user-supplied callback.

This is to generalize and support model variants like MESSAGE-Access and MESSAGE-Transport.

**PR checklist:**
- [x] Tests added
- [x] Documentation added

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -333,5 +333,5 @@ def test_solve_callback(test_mp, test_data_path):
         # Model iterates automatically
         scen.solve(callback=change_distance, **solve_args)
 
-    # Solution reached after 4 iterations, i.e. for f[4 - 1] == 90.0
+    # Solution reached after 4 iterations, i.e. for d[4 - 1] == 2.5
     assert scen.iteration == 4
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -331,7 +331,7 @@ def test_solve_callback(test_mp, test_data_path):
                'indefinitely unless True is returned.')
     with pytest.warns(UserWarning, match=message):
         # Model iterates automatically
-        scen.solve(**solve_args, callback=change_distance)
+        scen.solve(callback=change_distance, **solve_args)
 
     # Solution reached after 4 iterations, i.e. for f[4 - 1] == 90.0
     assert scen.iteration == 4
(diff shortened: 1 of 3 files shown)


gidden:
@MiguelPobleteCazenave would you mind taking a first pass in reviewing? Happy to do a second after the first round.

MiguelPobleteCazenave:
sure, i'll give it a look later today

khaeru:
> Wouldn't be desirable to pass more than just one argument for the callback (in case you want to have different versions of the other models, e.g., different levels of access targeted, different regions or in transport, different types of vehicles targeted)?

Ah, good call. I will add this now.

The user can also do this in a bunch of other ways, because the callback function will usually be defined in the same scope where `solve()` is called:

```python
myparam1 = 10.2

def cb(scenario):
    # Use a global variable
    global myparam1

    # Retrieve data stored with the callback
    myparam2 = cb.data['myparam2']

    # Read a file
    with open('myparam3.txt') as f:
        myparam3 = f.read()

    # do some calculations
    return convergence

# Store data with the callback after it has been defined
cb.data = {'myparam2': 23.5}

# Callback will have access to myparam[123]
scenario.solve(callback=cb)
```

khaeru:
@MiguelPobleteCazenave done

tests/test_core.py
```diff
@@ -269,3 +271,67 @@ def test_log_level(test_mp):
...
+        # 'iteration' variable stored on the Scenario object
+        set_d(scenario, d[scenario.iteration])
+
+        # commented: see below
+        # # Trigger another solution of the model
+        # return False
+
+    # Warning is raised because 'return False' is commented above, meaning
+    # user may have forgotten any return statement in the callback
+    message = (r'solve\(callback=...\) argument returned None; will loop '
```
gidden: This seems to be checking that the message is the same as that defined in `core.py`. Is it worth making this a top-level defined string there, e.g., `CALLBACK_WARNING=..` and here setting `message = core.CALLBACK_WARNING`? As is, it seems like if someone later edits the callback warning without touching this code as well, the test may not behave as expected.

gidden:
Per our in person conversation, we will leave test as is and let future devs deal with it if it becomes an issue.

PR #120, opened 2019-02-15
Clone scenario: check and fix (#109)

zikolach:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [ ] Documentation Added
- [ ] Description in RELEASE_NOTES.md Added

# Adding to RELEASE_NOTES.md

Please add a single line in the release notes similar to the following:

```
- (#XX)[http://link-to-pr.com] Added feature which does something
```

PR #127, opened 2019-03-06
Add unit tests to illustrate unit/region copy failure during `clone()`

danielhuppmann:


zikolach:
@danielhuppmann there is a small problem with cloning.

We have 2 methods: clone and commit. We assume that until we call commit - no changes are saved to the database.

That doesn't work like this for adding timeseries (and for clone as adding timeseries is part of it). Non existing units are added to DB on adding new timeseries, regions - not.

Would be good to discuss it as implementing saving regions the same (hacky) way it is done with units is tricky (we don't have parent/hierarchy inside addTimeseries). I would rather change it to keep correct assumption from first sentence (clone => in-memory, commit => DB).

danielhuppmann:
@zikolach, this cannot be correct - if I call `Scenrio.clone()`, the returned `Scenario` has been saved to the database and is not checked out...

zikolach:
@danielhuppmann yes, you are right, `commit` is called from Scenario constructor during cloning.

I meant a little bit different aspect - probably `clone` should not save changes to DB as e.g. other methods which change Scenario content do not do (adding parameters/variables/timeseries etc, except saving units). Therefore clone **should not** be different than e.g. creating new scenario, and user has to explicitly call `commit` after cloning to persist scenario instance.

danielhuppmann:
@zikolach, third comments here.

First, the core API has been in use now for two years by more than two dozen colleagues, so we should be conservative about changes - if we find critical issues, we can make improvements, but given that we have now a broad range of tutorials and workflow scripts implemented across multiple research groups (Clara in Munich, Alex in London, Pedro in Rio, Felipe in Santiago, students in Milan and at TU Wien), the default should be: no change.

Second, I don’t agree with the merit that this would make (a lot) more sense. There are two ways to get a scenario:
1) load an existing scenario from the database. As a start, a user may only want to view data (not edit anything), so it makes sense to not have that scenario checked out.
2) create a new scenario. As there is no data, it makes sense that this scenario is checked out from the start because this will be the first thing a user will do.

Now, cloning a scenario may be seen as having a starting point for a „new“ scenario (in which case your suggestion to do a `clone().commit()` makes sense) or as making a copy in the database and then viewing the copy (in which case the current implementation is the right way). I think that these two philosophies are here equally valid.

Third, the current implementation is consistent with a detailed changelog - you make a clone from an existing scenario, and you start by checking out the scenario. This way, we have a clear history. If we were to implement your proposal, we would either loose the information concerning which scenario a new scenario was cloned from and any changes done before the first commit, or we would have to rewrite the changelogging. Possible, of course, but do should not strive to make the code more complicated if the case for it is not string (see items 1 and 2).

zikolach:
@danielhuppmann I added new ixmp.jar to this PR built from https://github.com/iiasa/ixmp_source/pull/187. Seems like tests pass.

zikolach:
Just to link it - target branch is part of PR #120

danielhuppmann:
closing in favour of #129

PR #132, opened 2019-04-01
Document version and default version handling

### LINKED ISSUE #101: Document expected usage of `default` characteristic of scenarios, especially in use with `clone()`
Per our discussion in #97, it would be useful to have a documentation page that covers aspects of how versioning is done on model/scenario data. There are two main topics I see that could be covered at the moment.

First is the application of the `default` characteristic. How is this updated? What happens if there is no default version? How does this interact with `clone()`?

Second is the possible results from `clone()` with various arguments. E.g., what if I only give `model`, only `scenario` or both? What if model and scenario are the same as the copy-ee?

This could be pretty simple page enumerating all possible procedures with expected outcomes.

khaeru:
This documents how versions and default versions are handled for some API calls.

It also includes one functional change: if `Scenario.clone()` is called on a subclass of `Scenario`, the subclass constructor is called to create the return value. This allows the method to be called on, e.g. `message_ix.Scenario` without overloading.

Closes #101

**PR checklist:**
- [x] Tests added (N/A)
- [x] Documentation added

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 # Next Release
 
 - [#130](https://github.com/iiasa/ixmp/pull/130): Recognize `IXMP_DATA` environment variable for configuration and local databases.
-- [#129](https://github.com/iiasa/ixmp/pull/129): Fully implement `Scenario.clone()` across platforms (databases).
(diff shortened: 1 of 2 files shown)


khaeru:
@danielhuppmann @zikolach please let me know if this is inconsistent with anything going on on the Java side.

gidden:
OK LGTM!

PR #133, opened 2019-04-04
Updated pytest version requirement to >=3.9

zikolach:
As a minor change I hope it does not need to be included to changelog


khaeru:
For the record, this is because some people were reporting that `tmp_path_factory` could not be found when running tests. This was introduced in pytest 3.9.

PR #141, opened 2019-06-06
small check in dantzig example to be able to run this more than once …

gidden:
Small update to `ixmp.testing` to allow for running the dantzig example on already existing databases


danielhuppmann:
I added this check in the functions in `core.py` on the `ixmp` branch for the next release - I think it makes more sense to have it there.

gidden:
OK, closing

PR #142, opened 2019-06-07
Requirements for MESSAGEix release 1.2

danielhuppmann:

- [x] Tests Added
- [x] Documentation Added

# Description

This PR fully implements the feature to clone scenarios across platforms. The `clone()` and `remove_solution(keep_solution=False)` functions now also delete IAMC-style timeseries in a scenario (in the model horizon) to ensure consistency between the model solution (variables and equations) and derived reporting.

This PR also includes the changes in the unerlying Java core to suport the forthcoming release 1.2 of the `message_ix` package.


gidden:
I'm starting to try to take a look here @danielhuppmann . I assume if this is merged, we need to also merge iiasa/message_ix#205 to keep them in line?

ixmp/utils.py
```diff
@@ -34,6 +34,13 @@ def islistable(x):
     return isinstance(x, collections.Iterable) and not isstr(x)
 
 
+def isyear(y, s):
```
gidden: This function seems clunky and doesn't provide the same functionality/interface as the rest of our `is*()` functions which simply say `True` or `False`. I think we would save more in code understanding if the two blocks that use this simply expand to:

```
if y is not None and isinstanc(y, int):
    do_stuff()
elif y is not None:
    raise error about type
```

gidden: And just remove this line - I think in total we're saving like 2-4 lines by using this function at the moment. If we notice it gets used much more frequently, then we can reassess.

ixmp/core.py
```diff
@@ -299,7 +306,21 @@ def add_region_synomym(self, region, mapped_to):
...
+        if region not in list(_regions['region']):
+            self._jobj.addNodeSynonym(mapped_to, region)
+        else:
+            _logger_region_exists(_regions, region)
+
+
+def _logger_region_exists(_regions, r):
+    region = _regions.set_index('region').loc[r]
+    msg = 'region `{}` is already defined in the platform instance'
+    if region.mapped_to is not None:
```
gidden: I would prefer we use explicitly indexing here (i.e., region['mapped_to']) so we all know this is a column and not an element on some custom dataframe class. Same below!

ixmp/core.py
```diff
@@ -1174,19 +1198,24 @@ def clone(self, model=None, scenario=None, annotation=None,
...
+                             'with `keep_solution=True`!')
 
-        platform = self.platform if not platform else platform
-        model = self.model if not model else model
-        scenario = self.scenario if not scenario else scenario
+        platform = platform or self.platform
+        model = model or self.model
+        scenario = scenario or self.scenario
+        args = [platform._jobj, model, scenario, annotation, keep_solution]
+        if isyear(first_model_year, 'first_model_year'):
```
gidden: see my comment whwere `isyear()` is defined

ixmp/core.py
```diff
@@ -1230,24 +1259,33 @@ def read_sol_from_gdx(self, path, filename, comment=None,
...
         Raises
         ------
         ValueError
-            If Scenario has no solution.
+            If Scenario has no solution or if `first_model_year` is not `int`.
         """
         if self.has_solution():
             self.clear_cache()  # reset Python data cache
-            self._jobj.removeSolution()
+            if isyear(first_model_year, 'first_model_year'):
```
gidden: see my comment whwere `isyear()` is defined

ixmp/utils.py
gidden: I see now it's also used one time in the iiasa/message_ix#205, but my preference here remains.

ixmp/core.py
```diff
@@ -1174,19 +1198,24 @@ def clone(self, model=None, scenario=None, annotation=None,
                 ' release, please use `scenario`')
             scenario = kwargs.pop('scen')
 
-        first_model_year = first_model_year or 0
+        if keep_solution and first_model_year is not None:
+            raise ValueError('Use `keep_solution=False` when cloning with '
+                             '`first_model_year`!')
+
+        if platform is not None and not keep_solution:
+            raise ValueError('Cloning across platforms is only possible '
```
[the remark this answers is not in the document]
danielhuppmann: Because the Java backend is a mess and implementing all features would even take more time. So the current approach is the first least-bad option:
 - if you are cloning in the same platform, you can discard the solution on the fly or shift the first model year
 - you can create an exact duplicate in a difference platform

ixmp/core.py
danielhuppmann: done

ixmp/core.py
danielhuppmann: done

ixmp/core.py
danielhuppmann: done

ixmp/utils.py
danielhuppmann: changed to `check_year()` as discussed offline

ixmp/core.py
```diff
@@ -1150,16 +1172,18 @@ def clone(self, model=None, scenario=None, annotation=None,
...
         annotation : str, optional
-            Explanatory comment for the clone operation.
-        keep_solution : bool
-            If :py:const:`True`, include data from an existing model solution
-            in the clone.
-        first_model_year: int
-            If given, all time series data in the Scenario is omitted from the
-            clone for years from `first_model_year` onwards. Time series data
+            Explanatory comment for the clone commit message to the database.
+        keep_solution : bool, default True
```
khaeru: [The convention](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) is to do like `keep_solution : bool, optional`. The default is shown in the built docs automatically; [see here](https://message.iiasa.ac.at/projects/ixmp/en/latest/api/python_core.html#ixmp.Scenario.clone).

ixmp/core.py
```diff
@@ -1174,19 +1198,24 @@ def clone(self, model=None, scenario=None, annotation=None,
...
+        if platform is not None and not keep_solution:
+            raise ValueError('Cloning across platforms is only possible '
+                             'with `keep_solution=True`!')
 
-        platform = self.platform if not platform else platform
-        model = self.model if not model else model
-        scenario = self.scenario if not scenario else scenario
+        platform = platform or self.platform
+        model = model or self.model
+        scenario = scenario or self.scenario
```
khaeru: Nice! Cleaner.

ixmp/core.py
```diff
@@ -1230,24 +1259,33 @@ def read_sol_from_gdx(self, path, filename, comment=None,
...
-        database.
+        If ``has_solution() == True``, model solution data exists in the db.
         """
         return self._jobj.hasSolution()
 
-    def remove_solution(self):
-        """Delete the model solution.
+    def remove_solution(self, first_model_year=None):
+        """Delete the model solution (variables and equations) and timeseries
+        data marked as `meta=False` (see :meth:`TimeSeries.add_timeseries`).
```
khaeru: The summary line of the docstring should fit on one line.

ixmp/testing.py
```diff
@@ -134,10 +150,9 @@ def make_dantzig(mp, solve=False):
     mp.add_region('DantzigLand', 'country')
 
     # initialize a new (empty) instance of an `ixmp.Scenario`
-    model = 'canning problem'
-    scenario = 'standard'
     annot = "Dantzig's transportation problem for illustration and testing"
-    scen = Scenario(mp, model, scenario, version='new', annotation=annot)
+    args = models['dantzig'].copy()
+    scen = Scenario(mp, version='new', annotation=annot, **args)
```
khaeru: `args` and `copy()` unnecessary…just do ``Scenario(…, **models['dantzig'])``.

ixmp/core.py
danielhuppmann: Thanks!

ixmp/core.py
danielhuppmann: done

ixmp/core.py
danielhuppmann: done

danielhuppmann:
thanks @khaeru and @gidden for the review - all suggested changes implemented (and cloning across platforms should now *really*  work)

gidden:
Thanks @danielhuppmann !!

I'll leave the branch alive, but it should be deleted upon the actual release.

PR #146, opened 2019-06-14
 DO NOT MERGE: commit to remove unit checking (branch renamed)

gidden:
new reporting requires better unit support, which will require model cleanups and pint unit registry, for now this commit changes all units to '-' (for reporting testing). FYI @khaeru

### DIFF
--- a/ixmp/reporting/computations.py
+++ b/ixmp/reporting/computations.py
@@ -23,7 +23,8 @@ xr.set_options(keep_attrs=True)
 
 
 # Calculation
-def sum(quantity, weights, dimensions):
+# TODO: should we call this weighted sum?
+def sum(quantity, weights=None, dimensions=None):
     """Sum *quantity* over *dimensions*, with optional *weights*."""
     if weights is None:
         weights, w_total = 1, 1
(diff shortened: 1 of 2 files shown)


gidden:
now built on #149 @khaeru this is the branch I'm using to do the profiling work + the upstream message_ix branch

khaeru:
I'm now confused what the status of this is. The `drop` kwargs to the computations seem like they are in #150? Which leaves only the single-line deletion of all units?

If units in `message_ix` or specific global model scenarios are causing problems, then let's:

1. add real features to `ixmp` or `message_ix` that do things like pre-substitute units in a _generalized_ way, e.g.: https://github.com/iiasa/ixmp/blob/62d9f1ae4095355e2b1bcf31cdf59d7b999e7ac3/ixmp/reporting/utils.py#L17-L20 

…and then (2) _use_ those features to handle the brokenness of units in those models and/or scenarios, e.g.:
- https://github.com/iiasa/message_ix/blob/e2b85e90b72e6c46d7fd1e3d64e4a09eaf8abd00/message_ix/reporting/__init__.py#L17-L22
- https://github.com/iiasa/message_data/blob/126591fe26356ee6d5fe1049dc5a4df543069864/reporting/data/global.yaml

I would strongly prefer this to ‘temporary’ changes in `ixmp` that might ossify and cause trouble when we later try to remove them.

gidden:
For sure this is not meant at all to make it into any non-testing branch. It was only so I could actually run the reporting on the global model.

gidden:
this is no longer needed to run reporting via our testing script, though not entirely sure why =)

PR #147, opened 2019-06-14
Feature/reporting2 qty testing

gidden:
@khaeru this extends `ixmp.testing`  to include `assert_qty_equal` and `assert_qty_allclose`. some outstanding questions in TODOs. Shim removed, tests pass.

### DIFF
--- a/ixmp/reporting/utils.py
+++ b/ixmp/reporting/utils.py
@@ -211,7 +211,7 @@ class AttrSeries(pd.Series):
     _metadata = ['attrs']
 
     def __init__(self, *args, **kwargs):
-        if isinstance(args[0], xr.DataArray):
+        if hasattr(args[0], 'attrs'):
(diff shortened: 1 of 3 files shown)


ixmp/reporting/utils.py
```diff
@@ -211,16 +211,19 @@ class AttrSeries(pd.Series):
     _metadata = ['attrs']
 
     def __init__(self, *args, **kwargs):
-        if isinstance(args[0], xr.DataArray):
+        if 'attrs' in kwargs:
+            # Use provided attrs
+            attrs = kwargs.pop('attrs')
+        elif hasattr(args[0], 'attrs'):
```
gidden: not sure if this is the exact right way to do things, but I had a test fail because `attrs` was being passed in the constructor with `args[0]` was an `xr.DataArray`.

PR #148, opened 2019-06-14
DO NOT MERGE: show example where cloning fails when db is closed

gidden:
Here is an example where cloning still fails if the db is closed and reopened

cc @danielhuppmann @zikolach

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -4,7 +4,7 @@ from numpy import testing as npt
 import pandas.util.testing as pdt
 
 import ixmp
-from ixmp.testing import make_dantzig, TS_DF, HIST_DF
+from ixmp.testing import make_dantzig, models, TS_DF, HIST_DF
 
 TS_DF_CLEARED = TS_DF.copy()
 TS_DF_CLEARED.loc[0, 2005] = np.nan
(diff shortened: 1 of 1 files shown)


tests/test_integration.py
```diff
@@ -104,18 +104,28 @@ def test_multi_db_run(tmpdir, test_data_path):
...
+    info = models['dantzig']
+    scen2new = ixmp.Scenario(mp2new, info['model'], info['scenario'])
 
     # check that sets, variables and parameter were copied correctly
-    npt.assert_array_equal(scen1.set('i'), scen2.set('i'))
-    pdt.assert_frame_equal(scen1.par('d'), scen2.par('d'))
-    assert np.isclose(scen2.var('z')['lvl'], 153.675)
-    pdt.assert_frame_equal(scen1.var('x'), scen2.var('x'))
+    npt.assert_array_equal(scen1.set('i'), scen2new.set('i'))
+    pdt.assert_frame_equal(scen1.par('d'), scen2new.par('d'))
```
gidden: this assert fails with this commit

gidden:
```
>       pdt.assert_frame_equal(scen1.par('d'), scen2new.par('d'))
E       AssertionError: DataFrame are different
E       
E       DataFrame shape mismatch
E       [left]:  (6, 4)
E       [right]: (0, 4)
```

the `mp2new` scenario has no `par` values

danielhuppmann:
cherrypicked into #142

PR #154, opened 2019-06-23
Update README and documentation

khaeru:
Simplify README and expand the RTD docs; correct URLs for the 0.2/message_ix 1.2 release.

Preview: https://message.iiasa.ac.at/projects/ixmp/en/docs-readme/


khaeru:
Note that some of these links refer to the `/stable/` version of the MESSAGEix docs, which do not yet exist: they will once 1.2 is released and we designate that as the "stable" version on RTD.

PR #157, opened 2019-06-24
Release v0.2

khaeru:
Following https://github.com/iiasa/message_ix/wiki/Release-procedure


setup.py
```diff
@@ -30,46 +29,29 @@
...
-        'ixmp': 'ixmp',
-    }
-    entry_points = {
+LIB_FILES = [x.split('ixmp/')[-1] for x in glob.glob('ixmp/lib/*')]
+
+setup(
+    name='ixmp',
+    version=versioneer.get_version(),
+    description='ix modeling platform',
+    author='Daniel Huppmann, Matthew Gidden, Volker Krey, '
```
gidden: actually here, can we just `cat` the contents of `Authors.md` (or refer to it)? In other projects I work on, if you author a commit, you get to be on the author list. Some are a bit more strict (keeping lists per release), but I think that is too much overhead for us..

gidden:
@khaeru can you add a similar release procedure here with the "loose" differences =). I leave to your discretion where that should be (top level, docs, etc.).

gidden:
2 comments but otherwise LGTM!

setup.py
khaeru: There's an AUTHORS.md in message_ix but none in ixmp. See also https://stackoverflow.com/questions/9999829/how-to-specify-multiple-authors-emails-in-setup-py#10005265 —which is why I suggest using "IIASA Energy Program [and contributors]" in packaging/the copyright notice, and then having the full list in a file/the docs.

khaeru:
> @khaeru can you add a similar release procedure here with the "loose" differences =). I leave to your discretion where that should be (top level, docs, etc.).

https://github.com/iiasa/message_ix/wiki/Release-procedure for now, since we don't want to be kicking off builds with minor edits to the notes as we go.

gidden:
Sounds good - if it makes sense, may want to make a small issue to move this post release

PR #164, opened 2019-07-03
Fix adding timeseries with region synonyms

zikolach:
The issue with adding timeseries was fixed in the PR iiasa/ixmp_source#204. Attached jar file is build from that PR.

Note that `scope` for `test_mp` was changed to `function` - as to be able to use close_db without need to call open_db in all other tests. Also fixes `test_add_region_synonym` relying on data added in `test_add_region` (tests should be independent and able to be executed separately).


tests/test_feature_timeseries.py
```diff
@@ -103,6 +102,55 @@ def test_timeseries_edit(test_mp_props):
...
+                                 'variable': ['Testing'],
+                                 'unit': ['???'],
+                                 '2010': [23.7],
+                                 '2020': [23.8]})
+    scen.add_timeseries(df)
+    scen.commit('updating timeseries in IAMC format')
+
+    scen = ixmp.TimeSeries(mp, *args_all)
+    obs = scen.timeseries()
+    exp = pd.DataFrame.from_dict({'region': ['World', 'World'],
```
danielhuppmann: suggest to write `['World'] * 2`, same for other lines below

tests/test_feature_timeseries.py
```diff
@@ -103,6 +102,55 @@ def test_timeseries_edit(test_mp_props):
...
+                                 '2010': [22.7],
+                                 '2020': [23.7],
+                                 '2030': [24.7],
+                                 '2040': [25.7],
+                                 '2050': [25.8]})
+    scen.add_timeseries(df)
+    scen.commit('updating timeseries in IAMC format')
+
+    exp = pd.DataFrame.from_dict(
+        {'region': ['World', 'World', 'World', 'World', 'World', 'World'],
```
danielhuppmann: suggest to use ´['World'] * 6´

tests/test_feature_timeseries.py
```diff
@@ -103,6 +102,55 @@ def test_timeseries_edit(test_mp_props):
     npt.assert_array_almost_equal(df['value'], obs['value'])
 
 
+def test_timeseries_edit_iamc(test_mp_props):
```
danielhuppmann: why not use `test_mp` here?

zikolach: done

gidden:
@khaeru it looks like we have a PR failing from the R tests. Could you take a quick look?

khaeru:
Hi all,

I think this is related to a fragile thing I had to insert into the AppVeyor CI configuration; *mea culpa*. Without explaining (it's very 麻烦), the offending lines are in `ci/appveyor-install.R`:
https://github.com/iiasa/ixmp/blob/b62c81791db43866d424f440685aa07221a15049/ci/appveyor-install.R#L8-L11

Who is the assignee of this PR? @zikolach? Could you please try deleting those lines and see what happens?

zikolach:
@gidden AppVeyor builds are still failing. Could you please help with the problem?

khaeru:
> I removed lines @khaeru suggested

From the log, the R `mime` package seems to be installed from the Windows binary now, so this fix worked.

Now I see this unfamiliar error:
```
package 'IRkernel' successfully unpacked and MD5 sums checked
Error in IRkernel::installspec() : 
  jupyter-client has to be installed but "jupyter kernelspec --version" exited with code 1.
Execution halted
```
Investigating, will comment again.

zikolach:
> thanks @zikolach - I'm not sure how this test actually ensures that the synonym issue is solved?

I added synonym `test_mp.add_region_synomym('Hell', 'World')` and then timeseries using it `df = pd.DataFrame.from_dict({'region': ['Hell'],`

zikolach:
@danielhuppmann can we merge this PR and [next one](https://github.com/iiasa/ixmp/pull/167)?

PR #165, opened 2019-07-04
WIP: Add support for handling geodata

fonfon:
Implement adding/getting/removing geodata. Geodata here means meta-information about layers stored in a geoserver.

I'll test these changes a bit more with the `ixmp_server_integration` to import geodata for hotspots, and then remove the WIP.

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [ ] Release notes updated. Add a single line at the top of `RELEASE_NOTES.md` similar to…
  ```
  - [#999](https://github.com/iiasa/ixmp/pull/999) Description from above.
  ```
  …then, delete this sentence to confirm you've understood the instructions.

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -374,6 +374,7 @@ class TimeSeries(object):
 
     # Version of the TimeSeries
     version = None
+    _timespans = {}
 
     def __init__(self, mp, model, scenario, version=None, annotation=None):
         if not isinstance(mp, Platform):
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -15,6 +15,7 @@ Release 0.2.0 coincides with MESSAGEix release 1.2.0.
 
 ## All changes
 
+- [#165](https://github.com/iiasa/ixmp/pull/165): Add support for handling geodata.
```
khaeru: This should be under "Next Release", not under "v0.2.0 / All changes" (since 0.2.0 was already released).

ixmp/core.py
```diff
@@ -595,6 +596,101 @@ def remove_timeseries(self, df):
...
+            self._jobj.removeGeoData(values[0], values[1], values[2], years,
+                                     values[3])
+
+    def _get_timespans(self):
+        self._timespans = {}
+        timespan_cls = JClass('at.ac.iiasa.ixmp.objects.TimeSeries$TimeSpan')
+        for timespan in timespan_cls.values():
+            self._timespans[timespan.ordinal()] = timespan.name()
+
+    def _get_timespan_name(self, ordinal):
```
zikolach: would it make sense to change java API to use strings instead of enums if JPype doesn't provide convenient way to work with enums?

fonfon:
> It seems to be a bit of overhead converting inputs and outputs from java to python and back. Can we improve on that?

I'd love to have a more generic way of converting data. A library for at least the most common cases would be great, but if it doesn't exist I'm not sure whether it's worth for us creating one ourselves.. 
About strings instead of enum: yes, if it's not a drawback for the Java backend;

zikolach:
@peterkolp current implementation stores geoserver-related details as string (containing json). Later on if we figure out it is not sufficient (e.g. we need to query runs based on layer info etc) then we'll change the format and perform data migration.
So for now the protocol details are defined by the application (which attributes to save to json) - defined in Scenario Explorer (view part) and server integration script (write part).

peterkolp:
@zikolach @fonfon I don't get the idea - is this "geoserver meta data json string" supposed to end up in meta column?
Can you point me to the view/write parts which Nikolai mentioned?
Is there any database schema holding example (geoserver meta) data (maybe the hotspots ones)?

zikolach:
@peterkolp ixmp_test_sr15@gp3.iiasa.ac.at

```sql
select * from iamc_tsgeodata;
```

```csv
421225,2000,"{""url"":""https://ienecat.iiasa.ac.at:8743/geoserver/testNetCDF/wms"",""layer"":""R85_NOX__emiss_ind"",""format"":""image/gif"",""time"":null}"
421225,2080,"{""url"":""https://ienecat.iiasa.ac.at:8743/geoserver/testNetCDF/wms"",""layer"":""R85_NOX__emiss_ind"",""format"":""image/gif"",""time"":null}"
421225,2050,"{""url"":""https://ienecat.iiasa.ac.at:8743/geoserver/testNetCDF/wms"",""layer"":""R85_NOX__emiss_ind"",""format"":""image/gif"",""time"":null}"

```

That's probably not latest import bot but should work for you as demo.

PR #135, opened 2019-04-27
Test rixmp using 'testthat'

khaeru:
This PR renames retixmp to rixmp, and adds tests using the standard `testthat` R package. As well:

- R tests are run on all platforms, and on both Travis and Appveyor CI infrastructure.
- Tutorial tests (including Python tutorials) are re-enabled on Appveyor.
- Appveyor configuration is cleaned up.

Key pieces:

- `rixmp/source/tests/`
  - `testthat/test_core.R`: example file with minimal tests of core (Python) `ixmp` features via (R) `reticulate`.
  - `testthat.R`: standard file to invoke `testthat`.
- `tests/test_r.py`: add 1 new fixture and 3 new methods that, in turn, invoke the R tests:
  - `test_r_build_and_check()`: The standard pair of `R CMD build`, `R CMD check` (as run by e.g. CRAN or Travis [as described here](https://docs.travis-ci.com/user/languages/r/)). This is slow because numerous checks are performed in addition to the unit tests.
  - `test_r_testthat()`: run *only* the unit tests.

To test-drive *just* these new tests, run one of:
```
$ pytest -k 'r_build'
$ pytest -k 'r_testthat'
```

## To complete
- [x] **@adrivinca, @peterkolp**: please provide code (e.g. an example R script) that breaks retixmp. If you can convert this into a test (i.e. by editing `test_core.R`), please do. (I can show you how to PR that into my `khaeru:r-testing` branch; or, simply send me the code for the tests.) Or, I can create the test.
  - We will complete this PR without this test, then (if needed) copy the code/infrastructure to develop tests for reticulate-message_ix, or open a new issue if the message_ix work reveals things to be tested in retixmp.

**Anyone:** please help answer the following questions:
- [x] Q1: the `DESCRIPTION` files for rixmp and retixmp show versions of 0.0.0.9000 and 0.0.1, respectively. Should these not be the same as the ixmp Python package?
  - *A1:* updated to 0.1.3.9000. This will need to be set to 0.2.0 when the next release is made.
- [x] Q2: the existing method `test_run()` in `test_r.py` invokes the file `tests/r_tests.r`. I plan to convert this to a `testthat` test. What does the line `source(file.path(Sys.getenv("IXMP_R_PATH"), "ixmp.R"))` do? This file does not seem to exist in the repository.
  - *A2:* removed.
[...]

PR #170, opened 2019-07-18
CI

zikolach:
- update docker image for tests
- add teamcity test reporting


zikolach:
I setup a [build configuration](https://ene-builds.iiasa.ac.at/viewType.html?buildTypeId=IxmpCore_IxmpPython) on Teamcity as Travis/Appveyor builds are normally takes long time to finish. It checks only linux environment with python 3.7 for now.

khaeru:
> I setup a [build configuration](https://ene-builds.iiasa.ac.at/viewType.html?buildTypeId=IxmpCore_IxmpPython) on Teamcity

This asks me for a username and password. How do I obtain these?

setup.py
```diff
@@ -24,6 +24,7 @@
 ]
 
 EXTRAS_REQUIRE = {
+    'ci': ['teamcity-messages'],
```
khaeru: Is this used somewhere (e.g. `pip install .[ci]`) in the Docker build/Teamcity tests?

Dockerfile
```diff
@@ -1,10 +1,23 @@
...
-WORKDIR /
-ENV IXMP_PATH /ixmp
-ENV IXMP_R_PATH /ixmp/ixmp
-RUN pip2 install xlsxwriter xlrd
-RUN pip3 install xlsxwriter xlrd
-RUN cd /ixmp && python2 setup.py install 
-RUN cd /ixmp && python3 setup.py install
+FROM openjdk:11
+# installing python 3.7
+ENV PYTHON_VERSION=3.7.3
```
khaeru: As mentioned on Slack, the preference would be to invoke/re-use one of the scripts in the `ci/` directory to handle these steps.

This would mean that changes to the build/test environment would only need to be made in 2 places, rather than 3, and reduce the burden of maintenance.

khaeru:
The Travis failure seemed to be a time-out; restarted that job. As mentioned in Slack, I will look into the AppVeyor failures.

setup.py
zikolach: yes exactly https://ene-builds.iiasa.ac.at/admin/editRunType.html?id=buildType:IxmpCore_IxmpPython&runnerId=RUNNER_7&cameFromUrl=%2Fadmin%2FeditBuildRunners.html%3Fid%3DbuildType%253AIxmpCore_IxmpPython%26init%3D1&cameFromTitle=

Dockerfile
zikolach: I am not sure what exactly we may re-use. Can we discuss this and maybe make changes together?

khaeru:
> The Travis failure seemed to be a time-out; restarted that job.

Okay, Travis passes.

zikolach:
> Okay, Travis passes.

@khaeru do you know if we can do anything about build time on travis/circleci and appveyor. Waiting for an hour seems a bit to much? And also failure rate (timeouts and other issues) is quite high which makes it even worse in combination with long time

khaeru:
For posterity, the issue with AppVeyor seems to be related to the following.

On Windows, R tries to install binary (pre-compiled) packages; where these aren't available, it falls back to compiling from source. The problem is that, as minor/patch versions increment (unpredictably), then sometimes the binaries are briefly behind the source versions, and so AppVeyor tries (and fails) to compile the source packages.

I think what this needs is some way of forcing R to install only binaries; will investigate further.

```
also installing the dependencies 'sys', 'ellipsis', 'zeallot', 'ini', 'backports', 'ps', 'askpass', 'fansi', 'utf8', 'vctrs', 'clipr', 'clisymbols', 'curl', 'desc', 'fs', 'gh', 'glue', 'purrr', 'rlang', 'rprojroot', 'whisker', 'yaml', 'processx', 'R6', 'assertthat', 'mime', 'openssl', 'prettyunits', 'xopen', 'brew', 'commonmark', 'Rcpp', 'stringi', 'stringr', 'xml2', 'magrittr', 'praise', 'htmltools', 'pillar', 'base64enc', 'usethis', 'callr', 'cli', 'digest', 'git2r', 'httr', 'jsonlite', 'memoise', 'pkgbuild', 'pkgload', 'rcmdcheck', 'remotes', 'roxygen2', 'rstudioapi', 'sessioninfo', 'testthat', 'withr', 'repr', 'evaluate', 'IRdisplay', 'pbdZMQ', 'crayon', 'uuid'

  There are binary versions available but the source versions are later:
         binary source needs_compilation
processx  3.4.0  3.4.1              TRUE
openssl     1.4  1.4.1              TRUE
callr     3.3.0  3.3.1             FALSE

[snip]

Warning messages:
1: 
In install.packages(c("devtools", "IRkernel"), quiet = TRUE) :
 
 installation of package 'processx' had non-zero exit status
2: 
In install.packages(c("devtools", "IRkernel"), quiet = TRUE) :
  installation of package 'openssl' had non-zero exit status
3: 
In install.packages(c("devtools", "IRkernel"), quiet = TRUE) :
  installation of package 'callr' had non-zero exit status

[snip]

Rscript .\ci\appveyor-install.R 2
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called 'callr'
Calls: :: ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
Command exited with code 1
```

khaeru:
> @khaeru do you know if we can do anything about build time on travis/circleci and appveyor. Waiting for an hour seems a bit to much? And also failure rate (timeouts and other issues) is quite high which makes it even worse in combination with long time

We don't use CircleCI anymore. For the others, I agree it can be frustrating; e.g. for #135, I made over 170 commits and had to wait for CI to complete each time. For other PR's, it's not terrible, IMHO.

If that's a significant problem, we can probably pay in one or more ways:
1. Our staff time to set up and maintain our own build infrastructure, which might be faster. (Incidentally, for an apples-to-apples comparison, how long does the build of the Docker images take? Where does that happen?)
2. Subscription to the commercial versions of the CI services, that might be faster.
3. Our staff time to tweak & streamline the current CI configuration, e.g. using caching to avoid repeating slow installation steps.

@gidden may have further thoughts.

Dockerfile
khaeru: The sequence of `ci/env.sh`, `ci/travis-before_install.sh`, and `ci/travis-install.sh` performs the same tasks, namely:
- Sets environment variables for Python and GAMS versions (though note that Miniconda is used on Travis, whereas Python is build from source here; we should pick one and be consistent).
- `curl`s the files.
- Installs both.
- Sets up a conda environment.
- Installs graphviz (on Travis, only for macOS).

PR #173, opened 2019-07-23
Make AppVeyor CI more robust; support pandas 0.25.0

### LINKED ISSUE #169: errors when trying R installation
I got the following errors when trying our R installation instructions (which does not mention the expectation of such errors):

```
$ R CMD build .
* checking for file ‘./DESCRIPTION’ ... OK
* preparing ‘rixmp’:
* checking DESCRIPTION meta-information ... OK
Error in loadVignetteBuilder(pkgdir, TRUE) : 
  vignette builder 'knitr' not found
Execution halted
```

(fixed by installing `knitr`)

then

```
$ R CMD INSTALL *.tar.gz
* installing to library ‘/home/gidden/.local/R/library’
ERROR: dependency ‘reticulate’ is not available for package ‘rixmp’
* removing ‘/home/gidden/.local/R/library/rixmp’
```

(fixed by installing `reticulate`)

### LINKED ISSUE #171: R continuous integration is fragile
As noted on #170 and elsewhere:

On AppVeyor, i.e. Windows, R tries to install binary (pre-compiled) packages; where these aren't available, it falls back to compiling from source. The problem is that, as minor/patch versions increment (unpredictably), then sometimes the binaries are briefly behind the source versions, and so AppVeyor tries (and fails) to compile the source packages:
```
also installing the dependencies 'sys', 'ellipsis', 'zeallot', 'ini', 'backports', 'ps', 'askpass', 'fansi', 'utf8', 'vctrs', 'clipr', 'clisymbols', 'curl', 'desc', 'fs', 'gh', 'glue', 'purrr', 'rlang', 'rprojroot', 'whisker', 'yaml', 'processx', 'R6', 'assertthat', 'mime', 'openssl', 'prettyunits', 'xopen', 'brew', 'commonmark', 'Rcpp', 'stringi', 'stringr', 'xml2', 'magrittr', 'praise', 'htmltools', 'pillar', 'base64enc', 'usethis', 'callr', 'cli', 'digest', 'git2r', 'httr', 'jsonlite', 'memoise', 'pkgbuild', 'pkgload', 'rcmdcheck', 'remotes', 'roxygen2', 'rstudioapi', 'sessioninfo', 'testthat', 'withr', 'repr', 'evaluate', 'IRdisplay', 'pbdZMQ', 'crayon', 'uuid'

  There are binary versions available but the source versions are later:
         binary source needs_compilation
processx  3.4.0  3.4.1              TRUE
openssl     1.4  1.4.1              TRUE
callr     3.3.0  3.3.1             FALSE

[snip]

Warning messages:
1: 
In install.packages(c("devtools", "IRkernel"), quiet = TRUE) :
 
 installation of package 'processx' had non-zero exit status
2: 
In install.packages(c("devtools", "IRkernel"), quiet = TRUE) :
  installation of package 'openssl' had non-zero exit status
3: 
In install.packages(c("devtools", "IRkernel"), quiet = TRUE) :
  installation of package 'callr' had non-zero exit status

[snip]

Rscript .\ci\appveyor-install.R 2
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called 'callr'
Calls: :: ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
Command exited with code 1
```

The fix is to force R to install only binary packages; [the comments on this StackOverflow question](https://stackoverflow.com/questions/51289395/installing-packages-in-r-by-using-binary-instead-of-source-files-how-to-change) suggest a way to do this.

cc @zikolach

khaeru:
Closes #169, #171.

**PR checklist:**

- [x] ~Tests added.~ N/A
- [x] Release notes updated.

### DIFF
--- a/ci/appveyor-install.ps1
+++ b/ci/appveyor-install.ps1
@@ -60,7 +60,7 @@ Exec { conda remove --force --yes ixmp }
 Exec { conda info --all }
 
 # Install graphviz (for dask.visualize)
-Exec { choco install graphviz }
+Exec { choco install --no-progress graphviz }
 
 # Set up r-appveyor
 Bootstrap
--- a/ci/appveyor-install.ps1
+++ b/ci/appveyor-install.ps1
@@ -53,7 +53,7 @@ Exec { conda update --quiet --yes conda }
 # each about 150 MB. Enable Appveyor caching or tweak conda configuration to
 # speed up.
 Exec { conda install --channel conda-forge  --quiet --yes `
-       ixmp[tests] pytest coveralls pytest-cov }
+       ixmp[tests] "pytest>=3.9" coveralls pytest-cov }
 Exec { conda remove --force --yes ixmp }
 
 # Show information
--- a/ixmp/reporting/utils.py
+++ b/ixmp/reporting/utils.py
@@ -1,5 +1,5 @@
 import collections
-from copy import copy
+from copy import deepcopy
 
 from functools import partial, reduce
 from itertools import compress
@@ -7,6 +7,7 @@ import logging
 from operator import mul
 
 import pandas as pd
+from pandas.core.generic import NDFrame
 import pint
 import xarray as xr
 
@@ -257,7 +258,7 @@ class AttrSeries(pd.Series):
     """
 
     # normal properties
-    _metadata = ['attrs']
+    _metadata = ('attrs', )
 
     def __init__(self, *args, **kwargs):
         if 'attrs' in kwargs:
@@ -324,6 +325,18 @@ class AttrSeries(pd.Series):
     @property
     def _constructor(self):
         return AttrSeries
+
+    def __finalize__(self, other, method=None, **kwargs):
+        """Propagate metadata from other to self.
+
+        This is identical to the version in pandas, except deepcopy() is added
+        so that the 'attrs' OrderedDict is not double-referenced.
+        """
+        if isinstance(other, NDFrame):
+            for name in self._metadata:
+                object.__setattr__(self, name,
+                                   deepcopy(getattr(other, name, None)))
+        return self
 
 
 def data_for_quantity(ix_type, name, column, scenario, filters=None):
--- a/ixmp/reporting/utils.py
+++ b/ixmp/reporting/utils.py
@@ -1,4 +1,5 @@
 import collections
+from copy import copy
 
 from functools import partial, reduce
 from itertools import compress
@@ -133,7 +134,7 @@ def collect_units(*args):
             log.debug('assuming {} is unitless'.format(arg))
             arg.attrs['_unit'] = ureg.parse_units('')
 
-    return [arg.attrs['_unit'] for arg in args]
+    return [copy(arg.attrs['_unit']) for arg in args]
 
 
 # Mapping from raw -> preferred dimension names
--- a/ixmp/reporting/utils.py
+++ b/ixmp/reporting/utils.py
@@ -134,7 +134,7 @@ def collect_units(*args):
             log.debug('assuming {} is unitless'.format(arg))
             arg.attrs['_unit'] = ureg.parse_units('')
 
-    return [copy(arg.attrs['_unit']) for arg in args]
+    return [arg.attrs['_unit'] for arg in args]
 
 
 # Mapping from raw -> preferred dimension names
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,8 @@
 
 # Next Release
 
+- [#173](https://github.com/iiasa/ixmp/pull/173): Make AppVeyor CI more robust; support pandas 0.25.0.
+
 # v0.2.0
 
 Release 0.2.0 provides full support for Scenario.clone() across platforms (database instances), e.g. from a remote database to a local HSQL database.
--- a/ci/appveyor-install.R
+++ b/ci/appveyor-install.R
@@ -3,10 +3,16 @@ step <- commandArgs(TRUE)[1]
 options(repos=c('https://cloud.r-project.org'))
 
 if ( step == '1' ) {
+  # Install only binary packages on AppVeyor/Windows. This prevents build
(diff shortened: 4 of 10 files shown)


ci/appveyor-install.ps1
```diff
@@ -53,14 +53,14 @@ Exec { conda update --quiet --yes conda }
 # each about 150 MB. Enable Appveyor caching or tweak conda configuration to
 # speed up.
 Exec { conda install --channel conda-forge  --quiet --yes `
-       ixmp[tests] pytest coveralls pytest-cov }
+       ixmp[tests] "pytest>=3.9" coveralls pytest-cov }
```
khaeru: @gidden See [this log](https://ci.appveyor.com/project/gidden/ixmp/builds/26179568/job/o6hnn6rfkndwhifa). Conda was choosing to install pytest 3.2.2 on AppVeyor for Python 3 only (not Python 2.7). Shouldn't `ixmp[tests]` as a target here correspond to the 'tests' requirements in setup.py, which already specify >=3.9?

khaeru:
This rabbit hole goes deeper and deeper. I'm now seeing the following test failure that does not occur locally:
```
_____________________________ test_reporting_units _____________________________
    def test_reporting_units():
        """Test handling of units within Reporter computations."""
        r = Reporter()
    
        # Create some dummy data
        dims = dict(coords=['a b c'.split()], dims=['x'])
        r.add('energy:x',
              Quantity(xr.DataArray([1., 3, 8], attrs={'_unit': 'MJ'}, **dims)))
        r.add('time',
              Quantity(xr.DataArray([5., 6, 8], attrs={'_unit': 'hour'}, **dims)))
        r.add('efficiency',
              Quantity(xr.DataArray([0.9, 0.8, 0.95], **dims)))
    
        # Aggregation preserves units
        r.add('energy', (computations.sum, 'energy:x', None, ['x']))
        assert r.get('energy').attrs['_unit'] == ureg.parse_units('MJ')
    
        # Units are derived for a ratio of two quantities
        r.add('power', (computations.ratio, 'energy:x', 'time'))
        assert r.get('power').attrs['_unit'] == ureg.parse_units('MJ/hour')
    
        # Product of dimensioned and dimensionless quantities keeps the former
        r.add('energy2', (computations.product, 'energy:x', 'efficiency'))
>       assert r.get('energy2').attrs['_unit'] == ureg.parse_units('MJ')
E       AssertionError: assert <Unit('megajoule / hour')> == <Unit('megajoule')>
E         -<Unit('megajoule / hour')>
E         +<Unit('megajoule')>
tests/test_reporting.py:336: AssertionError
------------------------------ Captured log call -------------------------------
DEBUG    ixmp.reporting:__init__.py:262 Cull 5 -> 2 keys
DEBUG    ixmp.reporting:__init__.py:262 Cull 6 -> 3 keys
DEBUG    ixmp.reporting:__init__.py:262 Cull 7 -> 3 keys
DEBUG    ixmp.reporting.utils:utils.py:133 assuming x
                                           a    0.90
                                           b    0.80
                                           c    0.95
                                           dtype: float64 is unitless
```

khaeru:
The above failure occurs with pandas 0.25.0, but not with pandas 0.24.2.

khaeru:
> The above failure occurs with pandas 0.25.0, but not with pandas 0.24.2.

To elaborate, the following method in panda.core.generic.NDFrame (a parent class of pandas.Series) is responsible for transferring AttrSeries._metadata during some operations:
https://github.com/pandas-dev/pandas/blob/d7eb306cc1b6e3430895a7d2af0cfa1cbc8c4c06/pandas/core/generic.py#L5136-L5151

In pandas 0.25.0, for our ratio() reporting computation:
https://github.com/iiasa/ixmp/blob/a222fc8ab8cc9474241a486c603feddfd3c88d32/ixmp/reporting/computations.py#L88-L99

…this results in `result.attrs` being a reference to the same object as `numerator.attrs`, rather than a copy. So assigning `result.attrs['_unit']` also changes the units of `numerator`, affecting later calculations that use the same quantity.

The fix was to override `NDFrame.__finalize__`.

ixmp/reporting/utils.py
```diff
@@ -256,7 +258,7 @@ class AttrSeries(pd.Series):
     """
 
     # normal properties
-    _metadata = ['attrs']
+    _metadata = ('attrs', )
```
zikolach: out of curiosity - what is the idea behind replacing list with tuple?

ixmp/reporting/utils.py
```diff
@@ -324,6 +326,18 @@ def to_series(self):
...
+    def __finalize__(self, other, method=None, **kwargs):
+        """Propagate metadata from other to self.
+
+        This is identical to the version in pandas, except deepcopy() is added
+        so that the 'attrs' OrderedDict is not double-referenced.
+        """
+        if isinstance(other, NDFrame):
+            for name in self._metadata:
+                object.__setattr__(self, name,
+                                   deepcopy(getattr(other, name, None)))
```
zikolach: is it always needed to make a deep copy? may it have undesired consequences?

ixmp/reporting/utils.py
khaeru: This seems to be the practice in the upstream (pandas) code base. I guess it is so that no one accidentally mutates the sequence by adding/removing items.

ixmp/reporting/utils.py
khaeru: Currently `AttrSeries.attrs` is a `collections.OrderedDict`, and the only element is '_unit', which is always a `pint` Unit object.

If we started storing things like references to `ixmp.Platform` objects there, it would become a problem. But I think that would indicate a deeper design issue…

ixmp/reporting/utils.py
zikolach: https://github.com/pandas-dev/pandas/blob/master/pandas/core/generic.py#L155

Issue #158, opened 2019-06-25
Drop support for Python 2.7


khaeru:
See: https://python3statement.org. There will be no Python 2.8.

Parallel to iiasa/message_ix#177, and since ixmp 0.2 has been released, we should drop support for Python 2.7. This would include:

- Disabling CI for Python 2.7
- Removing code marked with `# py2 compat` or similar.

PR #175, opened 2019-07-25
Drop Python 2 CI

khaeru:
To partially address #158, CI for Python 2 is disabled on Travis & Appveyor.

**PR checklist:**

- [x] ~Tests added.~ N/A
- [x] ~Documentation added.~ N/A; will be written into the release notes for 1.0
- [x] Release notes updated.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,8 +3,6 @@
 
 environment:
   matrix:
-  - PYTHON_VERSION: "2.7"
-    PYTHON_ARCH: "64"
   - PYTHON_VERSION: "3.6"
     PYTHON_ARCH: "64"
   - PYTHON_VERSION: "3.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,8 @@ matrix:
   include:
   - os: linux
     env: PYENV=py37
-  - os: linux
-    env: PYENV=py27
   - os: osx
     env: PYENV=py37
-  - os: osx
-    env: PYENV=py27
 # turn these on once travis support gets a little better, see pyam for example
 #  - os: windows
 #    env: PYENV=py37
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ matrix:
 # turn these on once travis support gets a little better, see pyam for example
 #  - os: windows
 #    env: PYENV=py37
-#  - os: windows
-#    env: PYENV=py27
 
 r_packages:
 - IRkernel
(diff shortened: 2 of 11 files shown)


zikolach:
@khaeru how do we ensure (if we need it) user won't run it with python 2? Or maybe I missed something

khaeru:
> @khaeru how do we ensure (if we need it) user won't run it with python 2? Or maybe I missed something

Hm—well, as I said, we'll add it to the release notes. If they for some reason try to install and run/import it, it will fail immediately because the syntax will not be Python 2 compatible. Isn't that sufficient?

danielhuppmann:
@gidden would you mind taking a quick look and merge?

zikolach:
> > @khaeru how do we ensure (if we need it) user won't run it with python 2? Or maybe I missed something
> 
> Hm—well, as I said, we'll add it to the release notes. If they for some reason try to install and run/import it, it will fail immediately because the syntax will not be Python 2 compatible. Isn't that sufficient?

That's the thing - can (or should) we make a check for current version and return meaningful error message if it is less than 3? I am not sure what returned by python 2 discovered new syntax (except for the case with `class A(object)` when it uses some old "format" for classes instead of failing)

khaeru:
> > > @khaeru how do we ensure (if we need it) user won't run it with python 2? Or maybe I missed something
> > 
> > 
> > Hm—well, as I said, we'll add it to the release notes. If they for some reason try to install and run/import it, it will fail immediately because the syntax will not be Python 2 compatible. Isn't that sufficient?
> 
> That's the thing - can (or should) we make a check for current version and return meaningful error message if it is less than 3? I am not sure what returned by python 2 discovered new syntax (except for the case with `class A(object)` when it uses some old "format" for classes instead of failing)

Consider a file test.py:
```python
def foo(*args, args):
    pass
```
Then:
```
$ python2 test.py 
  File "test.py", line 1
    def foo(*args, arg):
                     ^
SyntaxError: invalid syntax
```

Even just to "check for current version and return meaningful error message", the interpreter has to parse the entire file. So *every* definition or import that is not py2-compatible would need to be protected, e.g. in test.py:
```python
import sys

if sys.version_info[0] == 2:
    print('Nope.')
    sys.exit(0)

from _test import foo

foo()
```
and then a new file must be created, _test.py:
```python
def foo(*args, args):
    pass
```

This:
1. doesn't accomplish anything, since the user is still unable to do anything;
2. defeats the point of dropping py2 support, which is to avoid weird stuff like this; and
3. doesn't seem to be common practice in other projects, from a casual glance around.

gidden:
thanks @khaeru !

PR #176, opened 2019-07-26
Update docker image for tests

zikolach:
- update docker image for tests
- add teamcity test reporting
- add missing dependency for python module (_lzma)

The idea behind this change is to get faster feedback from CI by:
- having a build configuration on teamcity (as opposite to travis/appveyor run tests on single platform as quick as possible)
- add tests reporting feature to teamcity build (show number of successful/failed tests on overview screen)
- make available docker image with all dependencies to run ixmp without need to install anything on host system (apart docker), e.g. use it as base to run server integration scripts


zikolach:
> Hi @zikolach. Given that our current CI already installs and runs tests successfully, I would suggest mapping the Dockerfile to use a much of that already-existing infrastructure. For example, we for sure do not need to build python itself from source - you can instead use miniconda and make sure everything is stable in separate environments. See e.g., the start of our ci installation [here](https://github.com/iiasa/ixmp/blob/master/.travis.yml#L36)

@gidden the idea of changing Dockerfile was to have working **base** docker image based on debian having up-to-date java and python (+ maybe gams). Nothing more (especially related to testing) as base image is intended to be used for production (e.g. running integration scripts as well). I am not sure if the way to install Python is the best one (just had troubles to install java on top of base official python image, so found this way working) and where Dockerfile for such an image should be located. I think we should discuss it offline/in person as there are other pieces in ci shell scripts (e.g. windows/macos support) which was most probably not intended to be used with docker.

gidden:
sounds good!

zikolach:
> the way to install Python is the best one (just had troubles to install java on top of base official python image, so found this way working)

@gidden I updated Dockerfile to use latest debian (buster) distro. Now python3 is installed from repository (not sources).

I am not sure I fully understand how dependency caching is set up, e.g. https://github.com/iiasa/ixmp/blob/master/ci/env.sh#L3 - seems should work for Travis, but does it work for Appveyor?

gidden:
hey @zikolach apologies for the delay here. I see that there is a `run_tests.sh`, but am not sure where/how these tests are being run. Do we need to add a teamcity hook on the repo?

zikolach:
@gidden no, there is a [build configuration on teamcity](https://ene-builds.iiasa.ac.at/admin/editRunType.html?id=buildType:IxmpCore_IxmpPython&runnerId=RUNNER_7&cameFromUrl=%2Fadmin%2FeditBuildRunners.html%3Fid%3DbuildType%253AIxmpCore_IxmpPython%26init%3D1&cameFromTitle=) which has build step running following script in container built from Dockerfile:
```
#!/bin/bash

source /opt/python3/bin/activate

pip install .[tests,ci]

py.test tests
```
Until this PR is merged I use a workaround (another build step containing exactly same Dockerfile content as this PR).
`run_tests.sh` is just for convenience to run tests locally (in docker) if one doesn't want to setup local python environment.

zikolach:
Usually webhooks are added from teamcity (there is UI for that), but with latest release either github plugin or something else got broken - so it works now without webhooks just polling once in a while repository for changes.

gidden:
ok, great. Is it possible to put these files not in the root directory? For example in the `ci` folder or (even better) `ci/teamcity`? We should also make specific folders for the other CIs..

setup.py
```diff
@@ -24,6 +24,7 @@
 ]
 
 EXTRAS_REQUIRE = {
+    'ci': ['teamcity-messages'],
```
khaeru: I would prefer to avoid this.

A similar example is `pytest-cov` or `coveralls`. These are not required for the end-user to run tests on the installed binary/source Python package; we only use them on CI. So the CI scripts install them, and they don't appear in setup.py.

Alternate locations would be in requirements.txt, or in run_tests.sh

run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
khaeru: Agree with @gidden this file should be in the ci/ directory.

Dockerfile
```diff
@@ -1,10 +1,17 @@
...
-RUN cd /ixmp && python3 setup.py install
+# installing python 3
+ARG VIRTUAL_ENV=/opt/python3
+RUN apt-get update && apt-get install -y --no-install-recommends curl unzip graphviz python3 python3-pip && \
+    pip3 install virtualenv && \
+    virtualenv -p python3 $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# installing GAMS
+ARG GAMS_VERSION=27.3.0
```
khaeru: We need to make it clear, somewhere, that when the GAMS version is incremented in the other CI scripts, it should also be done here, or vice versa. ci/env.sh currently uses GAMS 25.1.1.

At a minimum, comments in both places like `# See also ci/env.sh` and `# See also Dockerfile`.

zikolach:
Thanks for suggestions @gidden and @khaeru.

I don't fully agree as I treat Dockerfile to be used not only for tests/CI, but as delivery package/option. `run_tests.sh` is showing how to run tests if you have no python environment set up (but have docker one) - like having sample for `py.test tests` shell call in README (which is now removed from there).
Nevertheless I updated PR to get it finally merged.

khaeru:
> Dockerfile to be used not only for tests/CI, but as delivery package/option

This would be a nice additional option to have. However, if it were to truly be a first-class install option, then it would have to be added to the [installation page](https://message.iiasa.ac.at/projects/ixmp/en/stable/install.html) in the documentation—explaining when a user might want to choose it, other requirements, and then how to install. We would also need to consider whether it made best sense to put efforts into supplying a Docker image for ixmp, or instead maybe message_ix or even the entire MESSAGEix-GLOBIOM model, shipping with GAMS, input data, etc. etc.

Another PR that adds that would be welcome. The Dockerfile could be 'promoted' out of the ci/ directory at the same time.

Issue #119, opened 2019-02-13
Support non-GAMS models


khaeru:
`Scenario.solve()` ends with the following lines:
https://github.com/iiasa/ixmp/blob/aa7b2fee69b7c9a5b0c5736abf7dbf5cda55262b/ixmp/core.py#L1234-L1238

In order to support other, non-GAMS models, the following changes could be made:
1. Extend and generalize the code in `ixmp.model_settings`, adding:
   - a new `ixmp.model.Model` abstract class,
   - an `ixmp.model.GAMSModel` subclass.
2. Change `Scenario.to_gdx(…)` to `GAMSModel.write_data(scenario, …)`.
3. Change `run_gams(…)` to `GAMSModel.run(scenario, …)`.
4. Change `Scenario.read_sol_from_gdx(…)` to `GAMSModel.read_solution(scenario, …)`.

After this, users would be able to run non-GAMS models by subclassing `ixmp.Model()` and implementing the methods `write_data()`, `run()`, and `read_sol_from_gdx()`.

khaeru:
Here's a modified version of Dantzig's cannery/transport problem from the [Julia/JuMP optimization examples](https://github.com/JuliaOpt/JuMP.jl/tree/master/examples): [cannery.jl.txt](https://github.com/iiasa/ixmp/files/3356647/cannery.jl.txt)

To check that `Model` has been properly abstracted, a test could be written calling this through a `JuliaModel` class.

khaeru:
The branch [`feature/pyomo-model`](https://github.com/iiasa/ixmp/tree/feature/pyomo-model) contains some experimental code using [Pyomo](https://www.pyomo.org).

Issue #78, opened 2018-07-31
Database support


tzipperle:
Are there plans to support open-source relational database management systems (e.g. MySQL)?

danielhuppmann:
We currently have half an implementation using PostGreSQL, where half means that it should be fully operational but no-one has really used it in an actual workflow. However, you can open an HSQLDB-database using free software like the [DbVisualizer](https://www.dbvis.com)...

khaeru:
To slightly hijack this issue and add a few points:
- A similar but broader target is to decouple the Platform and Scenario classes from the ixmp_source/Java backend.
- With a `Backend` class defined, Platform/Scenario would depend on the Backend interface, but not care what happened behind it.
- The current ixmp_source (perhaps named “JDBCBackend” or similar) would continue to provide both remote (Oracle) and local (HSQLDB) connections.
- Other backends e.g. PostgresBackend, MySQLBackend, FileBackend, etc. could then be developed by anyone motivated to do so.
- This is related to #119, which is about the interface to models/solvers rather than databases/storage.
- FWIW HSQLDB is already an “open-source relational database management system” under the BSD license. It **is** true that it's not currently possible to use a **non-local** open-source RDBMS. To make use of ixmp based on 100% open-source software, other missing pieces might be:
  - GAMS and its solvers (again, see #119).
  - Some of the Java modules in ixmp_source; I'm not familiar.

zikolach:
@khaeru sorry for maybe outdated comment on your previous comment:

- decoupling is great idea (slightly different though - it could be something like direct (using ixmp jdbc DAO classes) and remote (using HTTP API)
- there should not be distinction on python side between different db implementations (as there is already such thing exist in java code)
- currently supported hsqldb, oracle and postgresql
- aim of python package is to provide flexibility to implement different models in addition to message and integration with python ecosystem (like tools for data analysis and visualization)
- support of different solvers might need to be implemented on java side (as e.g. persisting additional data structures in addition to gams ones (parameters, variables etc), timeseries data, layers references)

I hope it does make sense.

khaeru:
@tzipperle note that:
- with #182 merged, it is now nearly¹ possible to write a Backend storage class to support any kind of data storage.
- @gidden has a goal of doing this (or having his staff at Climate Analytics do it).
- #197 by @zikolach added support to the JDBCBackend for the free/open-source PostgreSQL database server; however, we do not yet have docs, examples, or tests for this. He can comment further on how easy these would be to set up.

[...]

PR #182, opened 2019-08-13
Add Backend, JDBCBackend, Model, GAMSModel classes

khaeru:
As described in [this comment](https://github.com/iiasa/ixmp/issues/78#issuecomment-519940541) on #78, this PR creates a `Backend` class separating the Python ixmp.core classes from the ixmp_source Java classes; and a `Model` class separating them from GAMS.

There is a corresponding PR iiasa/message_ix#249.

## How to review
- Look at [this diagram](http://www.plantuml.com/plantuml/uml/VLDDRnCn4BtFhnWj5q2AHQuLGarBLL4fW5ZkrcHtsbgixMZjZOgM_7USTRmsIvNRMFFUcpSVldQUUZ_e3WnfTdjg6DMBtkC8y5DfhhbNxF3ZdzKAwuODzSe2VE_8FzfUO_L1YFSLsf2jrXLcHW5Pn_mDDR_Oj03d86WAWG-VRsxdJF70rUhJP4GwE_GDYwHt4PJhNuIbHl0A6-ewXvhzisqTsaU0hMsvY-RkDzjwUWBCOHo5PkVeYQLw4TkxkjxStm7Cgg5QX5wXizs1MzpzBj9BWJpMfYFdS9xS26ymjI1tCO1ZWkLU0z3v7lMfvXIvfAYCnmFtJba3rBRI1uwRE6_VPQNPsNXcFxDcc0XxwWDowY8oQ1SWr7YCr6Hl126oyJHE8U29lHoCyg5WW1IcIv1tqWLSIvuYB3bwLIaRBuzdwInk87PawDIFb2hGfGmbnWG-OZ-OOXRboe-G_qHvFZWkD_NVHXQ33KBMCrPVzBvZpSPpYyhW0nseGdBuTVrliNmSfrhd0plT3PvZkTtS_lYst5zy9an-bOmZNBDf1ztz1G00).
- Read the rest of this comment
- Run the tests.
- Look at the preview of the revised docs. Do these clearly describe how the pieces come together?
  - Core API: https://ixmp-khaeru.readthedocs.io/en/feature-backend/api-python.html
  - New Backend API: https://ixmp-khaeru.readthedocs.io/en/feature-backend/api-backend.html
  - New Model API: https://ixmp-khaeru.readthedocs.io/en/feature-backend/api-model.html
- Skim the code (‘Files changed’ tab).

### Why do this?
Several reasons…

1. It more clearly separates three things that were tangled up in the current core.py code:
   1. The promised behaviour (API) of the current Backend, i.e. ixmp_source.
   2. Any conversions between Python and Java not already handled by JPype.
   3. Actual functionality and logic of the ixmp.core classes themselves.
2. Via (iii), the functionality of ixmp becomes more obvious to users.
2. It can be clear where (i) is performing voodoo (b/c not open source) that is not apparent from reading (iii).
   - This *could* (doesn't need to be) a guide or motivation to moving code to the Python side, where it is (a) more transparent and (b) possibly more easily maintained.
3. Via (i), the barrier is much lower for users to write other backends, as discussed in #78 and in Slack.

### Reading the code

- The current target here is to have **zero changes** in user-facing features; i.e., all tests pass, and no test has to be changed.
- There is a single Backend class that does not care very much about different classes of objects.
  - A hypothetical in-memory Backend might use a single object—an xarray.Dataset or even a simple dict()—and manipulate it in response to API calls.
  - The JDBCBackend **is** aware that ixmp_source contains many classes and manipulates them accordingly.
- *(Outdated; this was an intermediate step.)* I added some methods that spew noisy warnings, even though the tests pass. The warnings look like this:
  ```
  tests/test_core.py::test_init_scalar        
  tests/test_core.py::test_init_set                       
  tests/test_core.py::test_add_clone                                                                                                            
  tests/test_core.py::test_clone_edit                 
  tests/test_integration.py::test_run_clone           
  tests/test_integration.py::test_run_remove_solution
  tests/test_integration.py::test_multi_db_run                                                             
  tests/test_integration.py::test_multi_db_edit_source           
  tests/test_integration.py::test_multi_db_edit_target                                                                                         
    /home/khaeru/vc/iiasa/ixmp/ixmp/core.py:109: UserWarning: Accessing Platform._jobj in clone
  ```
  …and point to code that still needs to be refactored to work through the Backend. The PR will be ready once these warnings are all gone.
- *(Outdated; I've also done this in the current PR at the encouragement of @gidden, below)* The methods `to_gdx()` and `read_sol_from_gdx()` will, for now, force Backend to know what GAMS is. IMO this is less than ideal; per #119, in the future these could be moved to a `class GAMSModel(Model)`.

## PR checklist

- [x] Address remaining methods that access Java directly:
  - _jobj access in:
    - [x] Platform.add_region, add_region_synomym, add_unit, check_access, clone, regions, scenario_list
    - [x] TimeSeries.add_geodata, add_timeseries, check_out, commit, get_geodata, remove_geodata, remove_timeseries, timeseries
    - [x] Scenario.add_timeseries, check_out, clone, commit, get_meta, remove_solution, set_meta
  - [x] Scenario._item() calls in add_par, add_set, change_scalar, init_scalar, keys_for_quantity
  - [x] `ixmp.core._remove_ele`
- [x] Add 5 Backend methods required by `message_ix.core.Scenario`: ms_cat_set_elements, ms_cat_list, ms_cat_get_elements, ms_year_first_model, ms_years_active.
- [x] Delete ixmp.Scenario.years_active → MsgScenario.getTecActYrs; this was incorrectly located in **ixmp**.Scenario rather than message_ix.Scenario.
- [x] Add `Model`, `GAMSModel` classes, and transfer:
  - [x] Scenario.to_gdx, .read_sol_from_gdx.
  - [x] core.run_gams
- [x] Any further changes required for message_ix to work — all commits after 896b3b8.
- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


gidden:
Hey @khaeru, thanks for this monumental effort! I think this very strongly moves us in a good direction to a) have more available backends and b) start to detangle some of the MESSAGE-specific things on the java side. 

I also very strongly agree that the GAMS-specific things could be treated similarly. Do you think there is scope in this PR to do so? I can also help with that if desired.

In any case, I am happy to facilitate and review the PR (will add myself in a moment).

khaeru:
> I also very strongly agree that the GAMS-specific things could be treated similarly. Do you think there is scope in this PR to do so? I can also help with that if desired.

Sure, if there is help available then there is scope :smile:

My thoughts for the `Model` class design:

- Similar pattern:
  - Mostly empty file at `ixmp/model/__init__.py`.
  - Abstract base class `ixmp.model.base.BaseModel` with 2 methods:
    - `__init__(self, **kwargs)`: prepares the model. Keyword arguments vary by implementations.
    - `run(self, scenario)`: runs the model, including handling any data manipulation.
  - Concrete implementation `ixmp.model.gams.GAMSModel`:
    - `__init__()`: kwargs include the path to the GAMS files
    - `run()`:
      - Checks that `scenario.platform` is an instance of `JDBCBackend`, and raises an exception if not.
      - Does the equivalent of the current `Scenario.to_gdx()`, `ixmp.core.run_gams`, and `Scenario.read_sol_from_gdx()`.
      - Ideally, no Java type transformations here, either.
- Modify `JDBCBackend` with two new methods:
  - `s_write_gdx(self, s, path)`
  - `s_read_gdx(self, s, path)`
  - …so that these are clearly features of the `JDBCBackend` and not a general `Backend`.

khaeru:
In the future, this clarifies the design choices re: how/where file IO should be handled:
- if a GAMSModel were to work with a Backend other than JDBCBackend, then that Backend would _also_ need to be able to do GDX file I/O. But it might do so in Python. That code could then be used by both the OtherBackend and JDBCBackend.
- if a JDBCBackend were to work with a Model other than GAMSModel (e.g. JuliaModel) then that Model might need to write the Scenario to a file in a non-GDX format; and then read the solution from possibly another file in a different format. So again, the question of where this non-GDX I/O would live.

Other thoughts:
- A future `class MessageModel(GAMSModel)` can also advertise about what a “solution” comprises.
- Maybe it makes sense to have a GDXBackend? This is a half-formed idea, but the pattern would be kind of like:
  - clone a Scenario from JDBCBackend to a GDXBackend()
  - `GDXBackend.close_db()` ~~ current to_gdx()
  - model runs
  - `GDXBackend.open_db()` ~~ current read_sol_from_gdx()
  - clone or copy data back to JDBCBackend.

khaeru:
Updated the PR description somewhat to describe remaining work.

khaeru:
@gidden this is about done. Any interest in still reviewing it, or should I un-request and find someone else?

ixmp/backend/base.py
```diff
@@ -0,0 +1,803 @@
...
+    def s_clear_solution(self, s, from_year=None):
+        """Remove data associated with a model solution.
+
+        .. todo:: Document.
+        """
+
+    # Methods for message_ix.Scenario
+
+    @abstractmethod
+    def ms_cat_list(self, ms, name):
```
zikolach: I didn't get the idea of having different methods for scenarios (`s_`-prefixed) and message scenarios (`ms_`-prefixed). It contradicts with original idea of ixmp package to be generic and potentially support many models.

ixmp/backend/jdbc.py
```diff
@@ -0,0 +1,649 @@
...
+    'java.lang.Double',
+    'java.lang.Integer',
+    'java.math.BigDecimal',
+    'java.util.HashMap',
+    'java.util.LinkedHashMap',
+    'java.util.LinkedList',
+]
+
+
+class JDBCBackend(Backend):
```
zikolach: I would not expose directly any references to JDBC essence. It should be hidden for consumer. E.g. in some future we may implement (java side) some additional storage options (nosql, file-based etc) which is not related to JDBC, but expose exactly same API (to python/R consumers).

khaeru:
>  I think we should have a meeting where you can present your vision, we can draw some diagrams to have common understanding of existing and planned architecture.

Glad to do this. Who else besides us two should attend?

ixmp/backend/base.py
khaeru: > It contradicts with original idea of ixmp package to be generic and potentially support many models.

I agree. As I say in iiasa/message_ix#249, these are *stack level violations* where both ixmp and ixmp_source are currently *not* generic, but instead tailored to message_ix.

These were always there; this PR just makes it really obvious that they should not be there.

Our goal should be eventually to remove all `ms_` methods.
- I already dealt with firstmodelyear and years_active by moving the logic into message_ix. The correspond ixmp_source methods are no longer used.
- The category/mapping set features should be moved back into ixmp.
- Backend (code like) ixmp_source should not treat ixmp.Scenario or its subclasses differently. Any special treatment of message_ix.Scenario should be **in** message_ix.Scenario.

ixmp/backend/jdbc.py
khaeru: As mentioned in person, ixmp.Platform needs to choose and instantiate one Backend.
- For now, JDBCBackend is the only kind.
- Other Backends may provide storage without interacting with ixmp_source or any Java code.
- In the future, maybe, continued use of the "dbprops" and "dbtype" arguments could imply to use JDBCBackend, although I would prefer to clean up the semantics here.

khaeru:
Added a UML diagram to the PR description per discussions with @zikolach.

khaeru:
Per a long discussion with @gidden and @zikolach today, I will merge this after @zikolach's approving review, and then open further issues for follow-up tasks including:

- Reconsider mapping sets code location and logic.

  The methods and backend calls can probably be moved from message_ix.Scenario to ixmp.Scenario.
  Logic that creates certain "internal sets" with names like "map_*" in message_ix can be implemented in Python, based on the R code in ixmp_source.

- Implement writing to/from GDX in Python.

  This is necessary to have other Backend implementations that do not depend on the Java code in ixmp_source and to break the strict dependency between GAMSModel and JDBCBackend. It does not necessarily mean we will remove the toGDX method in ixmp_source, since that is used by other code.

- Move more message_ix initialization code from Java ixmp_source to Python.

ixmp/backend/base.py
zikolach: I see. Shouldn't then `s_` prefixes be removed? As by default methods will operate on scenario (abstract API). Isn't it clear from parameter list and/or docstrings? Maybe also mark `ms_` methods with deprecation comments as the code to be moved to `message_ix` package

khaeru: > Isn't it clear from parameter list

It should be, I guess, yes :thinking: Okay, I will make that change before merging, then.

> mark `ms_` methods with deprecation comments

I think this code might end up in ixmp.Scenario, in which case it is just a name change rather than a deprecation and removal. But if this ends up going out as-is in the upcoming release, then I will add notices to explain that.

PR #183, opened 2019-08-19
Fix typo in method name for `add-region-synonym`

zikolach:
Just a type in method name


behnam2015:
I tested and it's fine

PR #167, opened 2019-07-16
Check model access

zikolach:
Based on #164 (to be merged first)

Add core API method to check model access using auth server REST API.

PR also replaces and add a number of new java dependencies (such as jersey http client, jackson json parser etc) additionally upgrading jersey version to overcome incompatibilities with java 1.9+ versions.

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [ ] Release notes updated. Add a single line at the top of `RELEASE_NOTES.md` similar to…
  ```
  - [#999](https://github.com/iiasa/ixmp/pull/999) Description from above.
  ```
  …then, delete this sentence to confirm you've understood the instructions.

PR #185, opened 2019-08-21
Update ixmp.jar (timeseries changelog fix)

zikolach:
Update ixmp.jar based on commit b6812f206b03ed3f3a849cdf299a91c0230ffc26

- include timeseries changelog fix


khaeru:
@zikolach —as a general practice, is it possible for these PRs to include:
- a link to the last .jar file update, and
- a link that shows the changes in ixmp_source since then?

For instance, here it seems to be:
- #167 / 16f1fa0, and
- [this url](https://github.com/iiasa/ixmp_source/compare/2336941205f1c67f4fc66063dea7bf73ca22944b...b6812f206b03ed3f3a849cdf299a91c0230ffc26), i.e. the diff between iiasa/ixmp_source@2336941 and iiasa/ixmp_source@b6812f2

For those (i.e. me…) not actively involved in changes to the Java source code (many of which are for the scenario explorer rather than the message_ix stack), this makes it much faster to review.

zikolach:
@khaeru good idea. I added a bash which should do half-work for me https://github.com/iiasa/ixmp_source/pull/223/files. Why do you need link to last update? Do you know if there is a link showing file history like https://github.com/iiasa/ixmp/commits/5a7a45d643d212b4db54bb48f4547161d1acd238/ixmp/ixmp.jar, but having no specific commit?

khaeru:
> @khaeru good idea. I added a bash which should do half-work for me https://github.com/iiasa/ixmp_source/pull/223/files.

Awesome—thank you!

> Why do you need link to last update?

The first item lets us see changes in ixmp since the last sync; the second item, changes in ixmp_source. Problems could come from either side, or from conflicting changes on both sides. Having both links together makes it easier to think about this and quickly decide whether we need to do a closer investigation, or not.

> Do you know if there is a link showing file history like https://github.com/iiasa/ixmp/commits/5a7a45d643d212b4db54bb48f4547161d1acd238/ixmp/ixmp.jar, but having no specific commit?

I think you can give the branch name instead of the commit hash, e.g. https://github.com/iiasa/ixmp/commits/master/ixmp/ixmp.jar

Issue #166, opened 2019-07-10
Add Uniform Resource Identifiers (URIs) for Scenarios


khaeru:
Recent discussions at the MESSAGE weekly meeting have raised the (dis)advantages of having scenarios spread across multiple databases. This points to a need for a [Uniform Resource Identifier (URI)](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax) ([examples](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Examples)) for scenarios. One type of URI is a [Uniform Resource Name (URN)](https://en.wikipedia.org/wiki/Uniform_Resource_Name#Syntax) ([examples](https://en.wikipedia.org/wiki/Uniform_Resource_Name#Examples)).

For an ixmp (and thus message_ix) Scenario, the components of a unique identifier would include:
1. Database server, i.e. hostname, port, username, and password.
2. Database instance on the server **1**.
3. Model name.
4. Scenario name.
5. Version of the Scenario given by **3** and **4**.

So (purely for example, I have not thought about an ideal scheme), a URI might look like: `ixmp://ene_ixmp@gp3.iiasa.ac.at/CD_LINKS_SSP2_v2/baseline#43` or `urn:ixmp:gp3.iiasa.ac.at:CD_LINKS_SSP2_v2:variant_scenario:latest`.

This would simplify research discussions of specific scenarios, by making it easy to refer to a particular scenario in a totally unambiguous way.

More information:
- We see patterns like this in our code often: https://github.com/iiasa/ixmp/blob/b62c81791db43866d424f440685aa07221a15049/ixmp/testing.py#L30-L35 followed by: https://github.com/iiasa/ixmp/blob/b62c81791db43866d424f440685aa07221a15049/tests/test_integration.py#L115-L116
- There are also issues, e..g #136, related to ambiguity in scenario versions.
- As described at [URI references](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#URI_references), the full URI string would not always be needed. A particular `ixmp.Platform` instance is tied to components **1** and **2** from the above list, so loading a Scenario (e.g. `/CD_LINKS_SSP2_v2/baseline#43`) from that Platform implicitly supplies those components.
  - Components such as the database server username, password, and JDBC driver could be inferred or read from the user configuration.
- Currently the five components are specified in at least three different ways:
  - **1** and **2** via the `ixmp.Platform(…)` constructor *and* the optional database properties file.
[...]

PR #189, opened 2019-09-24
Add Uniform Resource Locators (URLs) for Scenarios

khaeru:
Implements #166 by providing utils.parse_url() and Scenario.from_url(). This expands on code that's been piloted via iiasa/message_data#41 and has proved useful.

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


khaeru:
@OFR-IIASA @behnam-zakeri here is the documentation for the 2 new methods, for help in review:
https://github.com/iiasa/ixmp/blob/f22e0f304947dcfd182ed8fc076da9a10069f86a/ixmp/utils.py#L53-L89
https://github.com/iiasa/ixmp/blob/f22e0f304947dcfd182ed8fc076da9a10069f86a/ixmp/core.py#L860-L884

behnam-zakeri:
Thanks @khaeru for this addition. I tested this on machine and the tests work and different combination of scenario identifiers work too (default or not, version number given or not, etc.). Just two points:
1. As discussed in persion, for getting connected to a central database through a `*.properties' file, that file needs to be under the default folder of ixmp database, i.e., `C:\Users\...\.local\share\ixmp`.
2. A minor question: the features loads the _Platform_ and a scenario instance together. As such, some of the methods of the _Platform_ cannot be used/assessed before making sure that the URL is correct for model/scenario combination too. 
For example, if I call a URL as `scen, mp = Scenario.from_url('ixmp://local/testtt/baseline')`, where the model name `test` is spelled wrongly, the call returns an error as this model is not in the database. However, the call doesn't return the Platfrom (`mp`) neither. So I'm not able to check the correct name of this scenario in the database by the _Platform_ method of `mp.scenario_list()`. Maybe I miss something out, or this is not an intended outcome.

khaeru:
> 2. For example, if I call a URL as `scen, mp = Scenario.from_url('ixmp://local/testtt/baseline')`, where the model name _test_ is spelled wrongly, the call returns an error as this model is not in the database. However, the call doesn't return the Platfrom (`mp`) neither. So I'm not able to check the correct name of this scenario in the database by the _Platform_ method of `mp.scenario_list()`. Maybe I miss something out, or this is not an intended outcome.

This is a useful point. IOW:
- In *script* use, not finding the model/scenario names is a failure, and the function should raise an exception.
- In *interactive* use (e.g. in a notebook), the user will want to play with the `mp` object. So `None` should be returned for `scen`, and only a warning (rather that exception) should be given.

Now I just need to decide which should be the default…

khaeru:
I will refresh this after #198 goes in.

Currently, it's a big ambiguous how the `PLATFORM` part of the URL relates especially to HSQLDBs that may be scattered around the user's machine. Either the database would have to be in a specific directory (limiting), or it might contain path separators (`/`), which would make the URL hard to read/interpret (defeats the purpose).

After #198, this becomes much simpler: `PLATFORM` will refer to the name of a platform config in the user's config.json file, i.e. previously set with `ixmp platform add [...]`.

khaeru:
@behnam-zakeri I have now updated this to use the new config system, so the PLATFORM part of a URL is always a configured platform name. As well, per your request I added an argument to from_url(). When loading the Scenario fails:
- errors='warn' (default) displays a warning but still returns (None, mp).
- errors='raise' raises the Exception to be caught by other Python code.

Please re-review when you have a moment!

behnam-zakeri:
@khaeru, thanks for improving this and adding the warning flag. I tested this and works as expected.

PR #205, opened 2019-11-13
Test utils.check_year and .pd_write

GamzeUnlu95:
As an exercise to write a couple of tests, I wrote these two test functions (test_check_year() and test_ispanda() ) for some part of the code which was not covered by tests. The tests address check_year() and pd_write() functions.

You can simply view the changes and run both tests to see if they work.


tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
+"""
+Test check_year(y,s) 
+
+@author: unlu
+"""
+from ixmp.utils import check_year
```
khaeru: Since we already have `test_utils.py`, can this be in that file?

tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
...
+import pytest
+
+"Enter a string value for y to raise a ValueError. - in this case test should be successfull"
```
khaeru: The docstring ("Enter a string…") for a function should be.
- Enclosed in triple quotes `"""`.
- Inside the function.
- A full sentence ending in a period.

tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
...
+ 
+def test_check_year():
+   
+    y1= "a"
```
khaeru: Ensure you use a linter; there should be a space on either side of the ` = ` assignment operator. See [PEP 8](https://www.python.org/dev/peps/pep-0008/#other-recommendations).

Actually, I'm not sure why Stickler is not running on your PR; I will investigate.

tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
...
+    s1= "a"
+      
+    with pytest.raises(ValueError):
+        assert check_year(y1,s1)
+    
+
```
khaeru: Linter will catch >1 blank lines inside a function.

tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
...
+def test_check_year():
+   
+    y1= "a"
+    s1= "a"
+      
+    with pytest.raises(ValueError):
+        assert check_year(y1,s1)
+    
+    
+    "If y = None."
```
khaeru: For inline comments, use the pound sign: `# If y = None`

tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
...
+    
+    assert check_year(y2,s2) == None
+        
+    "If y is integer."
+        
+    y3= 4
+    s3= 4
+  
+    assert check_year(y3,s3) == True
+
```
khaeru: Only 1 blank line at the end of a file.

tests/test_check_year.py
```diff
@@ -0,0 +1,38 @@
...
+"Enter a string value for y to raise a ValueError. - in this case test should be successfull"
+ 
+def test_check_year():
+   
+    y1= "a"
+    s1= "a"
+      
+    with pytest.raises(ValueError):
+        assert check_year(y1,s1)
```
khaeru: Linter will catch no space after `,`.

tests/test_utils.py
```diff
@@ -41,6 +41,14 @@ def test_pd_io_xlsx_multi():
         _obs = obs[k]
         pdt.assert_frame_equal(_obs, _exp)
 
+def test_ispanda():
```
khaeru: This test should:
- Probably be named `test_pd_write` since `pd_write` is the name of the function being tested.
- Use the [pytest `tmp_path` fixture](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture). As it is, you will notice that 'test.csv' is created in whichever directory you run `pytest`, which is not desirable; it confuses the user via git.

GamzeUnlu95:
Thanks for the comments @khaeru . I made revisions as you also suggested!

PR #209, opened 2019-11-15
Load timeseries list optimization

zikolach:
Replaces map (HashMap) with POJO as result type of getTimeseries method. Reduce memory consumption for large timeseries lists by approx. factor of 2 (see more iiasa/ixmp_source#233)


khaeru:
For posterity/anyone curious: “POJO” means "plain ol' Java objects” as opposed to more advanced Java classes.

PR #213, opened 2019-11-21
Add CachingBackend, address performance of JDBCBackend.item_get_elements

khaeru:
This PR:
- **Refactors JDBCBackend.item_get_elements() to *not* use pd.DataFrame.from_dict(),** which can be slow. Instead, an empty pd.DataFrame is allocated and values from the Java side are copied into it.
- **Re-adds `[:]` slice syntax when accessing arrays via JPype.** This was added in 2017 (554c9be) but removed in #182. It is significantly faster than accessing the object without the slice; see jpype-project/jpype#71. Although that issue is closed, this workaround still seems necessary. It is not discussed in the JPype documentation. This PR also adds a comment clearly stating this is necessary. Incorporates #214.
-  **Moves caching out of ixmp.core and into a CachingBackend class.** Other backends can inherit from this class and use its caching features; or just inherit from the base Backend.

In the longer run, we will need to add performance tests (e.g. using [pytest-benchmark](https://pytest-benchmark.readthedocs.io)) to ensure that the ixmp test suite *directly* tests that ixmp has satisfactory performance for data on the scale of MESSAGE-GLOBIOM (~10 dimensions, ~10⁶ rows). That will be a separate issue.

## How to review

- Read the code!
- Try to load large parameters and ensure that performance is better when the cache is hit.
  - Import the code and use a custom script to load some large MESSAGE-GLOBIOM parameters.
  - Change N_dim in `test_report_size` to 8 and run only that test with `pytest -k report_size`.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,7 +20,8 @@ Configuration for ixmp and its storage backends has been streamlined.
 
 - [#189](https://github.com/iiasa/ixmp/pull/189): Identify and load Scenarios using URLs.
 - [#182](https://github.com/iiasa/ixmp/pull/182),
-  [#200](https://github.com/iiasa/ixmp/pull/200): Add new Backend, Model APIs and JDBCBackend, GAMSModel classes.
+  [#200](https://github.com/iiasa/ixmp/pull/200),
+  [#213](https://github.com/iiasa/ixmp/pull/213): Add new Backend, Model APIs and CachingBackend, JDBCBackend, GAMSModel classes.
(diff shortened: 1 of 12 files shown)


khaeru:
@OFR-IIASA notes that this change brings the time to load the `land_out` MESSAGE-GLOBIOM parameter down from 470 seconds to 50 seconds, which is “reasonable”. This variable has about 1.9 million rows, so the benchmark is roughly 26 µs/row.

He also reports that the caching appears to work.

Issue #218, opened 2019-11-22
ixmp should set the same datatypes for both var and par 


OFR-IIASA:
when retrieving data, the year values for a parameter are returned as integers, whereas for variables these are returned as strings.

A quick test:

model = <model name of a scenario with a solution>
scen_name = <name of scenario with solution>
scen = message_ix.Scenario(mp, model, scen_name)
print(scen.par('relation_activity').year_act.unique().tolist())
print(scen.var('ACT').year_act.unique().tolist())

using this on a scenario results in:
print(scen.par('relation_activity').year_act.unique().tolist())
[1990, 1995, 2000, 2005, 2010, 2020, 2030, 2040, 2050, 2060, 2070, 2080, 2090, 2100, 2110]
print(scen.var('ACT').year_act.unique().tolist())
['1990', '1995', '2000', '2005', '2010', '2020', '2030', '2040', '2050', '2060', '2070', '2080', '2090', '2100', '2110', '1950', '1955', '1960', '1965', '1970', '1975', '1980', '1985']

I would guess that the issue is a result of:
ixmp.backend.jdbc.item_get_elements()

khaeru:
This is the same as iiasa/message_ix#268.

khaeru:
Closing this as “wontfix” in *ixmp*; the fix is in iiasa/message_ix#269.

Issue #215, opened 2019-11-22
Add performance tests


khaeru:
Use [pytest-benchmark](https://pytest-benchmark.readthedocs.io) or similar tools to test performance metrics.

The ixmp unit tests use simple models like the Dantzig cannery/transport problem to quickly test basic features. However, ixmp is also used (via message_ix) for MESSAGE-GLOBIOM, in which parameters have up to 14 indices and ~2 million rows. Speed and memory performance is important for these data.

The metrics should include the following:
- **Read time per row.** While testing #213, @OFR-IIASA stated that 50 seconds to load the `land_out` parameter with ~1.94 million rows is “reasonable”, while 470 seconds is not. These numbers correspond to 26 and 242 µs / row, respectively.
- **Write time per row.** In #66, @OFR-IIASA  reported a time of 40 minutes for certain data as unacceptable; but the size of the data was not stated.

Miscellanea:
- It may be difficult to test performance issues due to network speeds and the Oracle database server used by IIASA ENE. e.g. @OFR-IIASA reports that TimeSeries.commit() can be extremely slow in some cases.

khaeru:
cf. #227 and #298 for memory performance.

PR #219, opened 2019-11-22
Parameterize treatment of item dimensions' types (#218)

zikolach:
- bring back special handling for year dim
- extend to any dim containing year reference
- simplify df creation by using again from_dict

Note: it turns out from_dict has same/better performance in comparison with alternative aproach passing index to empty df upon creation.
It also allows to have int dtype (which empty df doesn't support because of NaNs)


zikolach:
@khaeru one of the test is failing because I changed backend method to let using url for hsqldb. I needed that to use in-memory db in test. What was your idea behind limiting hsqldb to use only file-based Dbs?

zikolach:
@khaeru thank you very much for reviewing and commenting on PR. Sorry for not clarifying it in first place - I should not probably request "normal" review, but rather ask for suggestions as I consider this changes as draft.

> Remove type conversion for years...

I commented on that in linked ticket https://github.com/iiasa/message_ix/issues/268#issuecomment-557879877. I am ok with both variants mentioned there. The change made here as appropriate ticket #218 exists in this repo. I will remove it from code for the time being, but let's discuss it quickly next week.

> Place the methods save_par and read_par inside their corresponding test functions. Then you can even move lines line scen = ixmp.Scenario(… out of save_par into the surrounding function test_read_par_10000—since that line is not what is being benchmarked.

Under normal circumstances your suggestion totally makes sense. But in current setup benchmark does execute the method multiple times (calculating min/max/mean etc), therefore it needs every time new version of scenario (or new database, or other aproach). Creating new scenario every time add constant time and should not have significant effect on large data volumes. Can you suggest how to implement it better with `pytest-benchmark`?
I'll put test to subfolder, but I would put it to separate file as non-functional, performance test.

> Consider the how in-memory hsqldb...

My intention was not to expose this "feature" to external usage, but rather make it possible to use for tests. Should we document features which solely used for testing?
My comment regarding "limiting" was referring to the fact that new implementation is restrictive in this particular aspect (not being said it is good or bad). Sorry for any possible misinterpretation.
If we are going to "advertise" this feature to be used by package users - let's make a separate issue/PR adding documentation/tests/etc.

Originally I didn't want to introduce a lot of changes, but since @OFR-IIASA experienced issues with performance reading parameters elements I got an impression ticket #218 is critical to fix rather quickly.

khaeru:
> it needs every time new version of scenario (or new database

Okay, that's a good point—I would then suggest to add a comment like `# Create a new, empty Scenario for the test` above that line.

> I'll put test to subfolder, but I would put it to separate file as non-functional, performance test.

Sure, maybe something like tests/backend/test_jdbc_performance.py. Then other backends could have a similar split.

> Should we document features which solely used for testing?

Yes, absolutely. Our test suite is already large and complex and I'm not sure anyone fully understands it. To give people in the future any hope of maintaining it (or allow other people, e.g. colleagues of @gidden, to contribute easily) we need to make it easy to read. I think that should include not making use of 'hidden' features. So if the feature is documented well enough to read the test suite, it is not much additional work to make it a public feature. (Our new RAs are a resource here: if they can't understand the feature used in testing, that means we can do better.) 

I understand that the message_data performance issue @OFR-IIASA encountered was urgent to fix; however, in our haste to respond, #213 was merged before we had time to think about it carefully. That in turn spawned #218, this PR, iiasa/message_ix#268, and iiasa/message_ix#269.

Since the performance of #213 was ruled “acceptable”, I think the additional performance improvement from this PR (BTW, see above where I said “leave a comment here with the benchmark values prior to this PR”) does not need to be similarly rushed.

zikolach:
> Add comments inline indicating...

Can we discuss it?

zikolach:
Sample output of jdbc backend performance test.
```
-------------------------------------------------------------------------------------- benchmark: 2 tests -------------------------------------------------------------------------------------
Name (time in ms)             Min                 Max                Mean             StdDev              Median                 IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_save_par_1000        74.7618 (1.0)      102.1507 (1.0)       87.6907 (1.0)       8.6063 (1.0)       89.6419 (1.0)       12.8233 (1.0)           5;0  11.4037 (1.0)          13           1
test_read_par_100000     690.2758 (9.23)     875.4846 (8.57)     786.7553 (8.97)     82.7296 (9.61)     756.9678 (8.44)     144.6853 (11.28)         3;0   1.2710 (0.11)          5           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
It indicated roughly:
- saving 1000-element parameter takes 87 milliseconds
- reading 100000-element parameter takes 786 milliseconds
The results are highly environment dependent and should not be used in combination with results taken from another environment.

zikolach:
As discussed with @khaeru we close this PR for the time being as:
- moving conversion to non-string dtypes to ixmp core is delayed until figuring out all the requirements (possibly for another typed of backends)
- in-memory hsqldb support would be addressed as separate ticket (#224)
- performance tests will be added separately (#215)

PR #220, opened 2019-11-24
Remove special handling for dimensions indexed by a 'year' set

khaeru:
#213 mistakenly added code in `ixmp.Scenario.par()` with special treatment for dimensions indexed by a set named 'year'. Per iiasa/message_ix#268 / #218, this is the **wrong place** for such code. It belongs in message_ix. 

*Mea culpa*! The change was too hurried.

See iiasa/message_ix#269, which adds this special handling in the correct place.

## How to review

Check that tests pass.

## PR checklist

- ~[ ] Tests added.~ N/A
- ~[ ] Documentation added.~ N/A
- ~[ ] Release notes updated.~ N/A; adjustment to a previous PR.

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -925,17 +925,7 @@ class Scenario(TimeSeries):
         filters : dict
             index names mapped list of index set elements
         """
-        result = self._backend('item_get_elements', 'par', name, filters)
-
-        # FIXME message_ix requires 'year' columns to be returned as integers
-        #       This code should be in a message_ix override of this method.
-        dtypes = {}
-        for idx_set, col_name in zip(self.idx_sets(name),
-                                     self.idx_names(name)):
-            if idx_set == 'year':
-                dtypes[col_name] = int
-
-        return result.astype(dtypes) if len(dtypes) else result
+        return self._backend('item_get_elements', 'par', name, filters)
 
     def add_par(self, name, key_or_data=None, value=None, unit=None,
                 comment=None, key=None, val=None):


khaeru:
FYI @OFR-IIASA.

ixmp/core.py
```diff
@@ -925,17 +925,7 @@ def par(self, name, filters=None, **kwargs):
...
-        # FIXME message_ix requires 'year' columns to be returned as integers
-        #       This code should be in a message_ix override of this method.
-        dtypes = {}
-        for idx_set, col_name in zip(self.idx_sets(name),
-                                     self.idx_names(name)):
-            if idx_set == 'year':
-                dtypes[col_name] = int
-
-        return result.astype(dtypes) if len(dtypes) else result
+        return self._backend('item_get_elements', 'par', name, filters)
```
zikolach: nit: Should `kwargs` be passed to backend's `item_get_elements` method?

khaeru: I think these `kwargs` should be removed entirely; they exist only to be forgiving of sloppy user code.

khaeru:
Thanks for the review!

> And please have a look and comment if such a parameterization of `item_get_elements` makes sense https://github.com/iiasa/ixmp/pull/219/files#diff-10a41276f3b0d79aaf6b26709015e05fR504

Per our discussion today, this will remain on the message_ix side for now in order to keep ixmp small and the Backend interface simple. At some point—when iiasa/message_ix#254 is addressed—we can set up a general mechanism for ixmp to respect explicit types for elements of different sets.

PR #217, opened 2019-11-22
Allow filters with non-members of index sets

khaeru:
Closes #216.

## How to review

- Check that tests pass.
- Read the code.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

PR #225, opened 2019-11-25
Apply as_str_list() to filters; test

### LINKED ISSUE #226: retrieving a parameter does not accept integers for for filtering years only strings
retrieving data with set filters should work when filtering for the year, provided as an integer.

In the first call, a parameter is retrieved with the year_act set to a integer: this does not return any results.  

> (Pdb) scen.par(par, filters={'node_loc': ['R11_CPA'], 'technology': ['bio_istig'], 'year_act': [2020]})
> Empty DataFrame
> Columns: [node_loc, technology, year_act, time, value, unit]
> Index: []

In the second call, the parameter is retrieved with the year_act set to a string: 

> (Pdb) scen.par(par, filters={'node_loc': ['R11_CPA'], 'technology': ['bio_istig'], 'year_act': ['2020']})
>   node_loc technology  year_act  time  value unit
> 0  R11_CPA  bio_istig      2020  year    0.4  GWa

khaeru:
FYI @zikolach @OFR-IIASA.

`filters={...}` arguments to methods like `Scenario.par(...)` accept any type of element that can be passed to string. This behaviour was previously **untested**; this PR adds tests and ensures the filters are always converted to `str`.

Closes #226.
Improves code added in #213, #217.

In particular, note these former lines: https://github.com/iiasa/ixmp/blob/e6b3c0d663b97b09a6ccec331ede1e2def992ca6/ixmp/backend/jdbc.py#L524-L531

## Details

Where the `values` (from the user's `filters`) contained, for example, `int(42)`, but the `elements` (from the Backend itself) were always `str`, e.g. `'42'`, this comparison would fail, and `filtered_elements` would be empty.

The added call to `as_str_list()` further up in the method ensures this is always a string-to-string comparison.

## How to review

Test with any downstream code, e.g. message_data.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


khaeru:
In c4743ec, `pytest -k test_filter_str` fails.

khaeru:
> New test passes

Turns out it does not pass on Travis [#765.1](https://travis-ci.org/iiasa/ixmp/jobs/616702748):
```
_______________________________ test_filter_str ________________________________
test_mp = <ixmp.core.Platform object at 0x7fe5e5e11c10>
    def test_filter_str(test_mp):
        scen = ixmp.Scenario(test_mp, 'model', 'scenario', version='new')

        elements = ['foo', 42, object()]
        expected = list(map(str, elements))

        scen.init_set('s')

        # Set elements can be added which are not str
        scen.add_set('s', elements)

        # Elements are stored and returned as str
        assert expected == scen.set('s').tolist()

        # Parameter defined over 's'
        p = pd.DataFrame.from_records(zip(elements, [1., 2., 3.]),
                                      columns=['s', 'value'])
        p_exp = p.astype({'s': str})

        scen.init_par('p', ['s'])
        scen.add_par('p', p)

        # Values can be retrieved using non-string filters
        exp = p_exp.loc[1:, :].reset_index(drop=True)
        obs = scen.par('p', filters={'s': elements[1:]})
>       pdt.assert_frame_equal(exp[['s', 'value']], obs[['s', 'value']])
E       AssertionError: Attributes are different
E
E       Attribute "dtype" are different
E       [left]:  float64
E       [right]: float32
tests/test_core.py:309: AssertionError
```

@zikolach @OFR-IIASA if one of you would like to adjust, please go ahead. Otherwise I can deal with this after the reporting meeting tomorrow.

khaeru:
@zikolach on further investigation, `test_cache_non_hashable` was checking for an exception message:
```
Object of type object is not JSON serializable
```
However, **only** on AppVeyor/Windows and Python 3.6, this was the following (note the quotes):
```
Object of type 'object' is not JSON serializable
```

657aa51 attempts to fix. Please re-review and merge if checks all pass; otherwise, please investigate.

Issue #203, opened 2019-11-11
.gms files on network drive not recognized


Jihoon:
Path setups for gams files from `gams.py` do not correctly bring in files on network drives.

The absolute paths from Path.resolve() cannot be recognized from GAMS for some reason. And it complains as follows:

`'\\hdrive\home$\u045\min\mydocuments\MESSAGE\message_ix\message_ix\model'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.`

Removing `.resolve()` seems to resolve the issue.

khaeru:
Thanks @Jihoon for this issue report.

> Removing `.resolve()` seems to resolve the issue.

[`pathlib.Path.resolve()` is used in several places](https://github.com/iiasa/ixmp/search?q=resolve&unscoped_q=resolve). Can you specify which one(s) you removed to fix this issue on your system?

Jihoon:
There are three uses I find in the `main` definition. I removed all three.

khaeru:
@Jihoon if that has solved the problem for you, can you please create a PR so that it can be resolved for everyone?

Jihoon:
> @Jihoon if that has solved the problem for you, can you please create a PR so that it can be resolved for everyone?

Thanks. Done #230.

PR #230, opened 2019-12-03
Fixing how ixmp locates GAMS files on network drives

Jihoon:
IXMP wasn't able to locate GAMS files on network drives, because it was getting absolute paths by calling `pathlib.Path.resolve()`, which cannot be handled for some reasons. By removing those `.resolve()` calls, I can avoid the errors. (#203)

## PR checklist

- [x] Tests added.
- [x] Release notes updated.


khaeru:
Thanks @Jihoon.

Thinking aloud: [here is the documentation for resolve()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve). Its main purpose is to do a few things:
- If the user gives something like `./path/foo/bar/../baz/file.gams`, it simplifies to `C:/current_directory/path/foo/baz/file.gams`, i.e. the special relative paths `.` and `..` are handled.
- If there is a symlink (on Linux/macOS), like `path/foo` points to `path/bar`, then `path/foo/file.gms` is converted to `path/bar/file.gms`.
- I think the special path `~` is also expanded to `$HOME`, the user's home directory.

With your changes, none of these things happen anymore. I *guess* that is okay, but read on.

Looking at the PR checklist.

> - [ ] Tests added.

- I don't think it's possible to add a test for Windows. We could add a test that would work on just your computer, but the [AppVeyor service we use](https://www.appveyor.com/docs/build-environment/) does not seem to have network drives, so it would not be possible to run that test there.
- We don't have a test for the case where the user gives the model files with a weird path like `path/foo/bar../../baz/../something/file.gms`. So this may break. However, I don't think anyone currently is using it this way. If they do, they can file a new issue and we can make a specific test based on their example.

So please strike this out.

> - [ ] Documentation added.

Probably not needed, so put a strike through: `- [x] ~Documentation added.~ N/A`

> - [ ] Release notes updated.

Please do this.

khaeru:
> Looking at the PR checklist.

I've addressed these.

PR #231, opened 2019-12-05
Improving ixmp reporting documentation

francescolovat:
Corrected typos and performed some formatting to the `ixmp` reporting documentation files in `/reporting`.

This has been done by also editing the _docstrings_ of different methods of the `Reporter` class as well as other main classes.

Please @khaeru go through this small modification when you'll have time so this PR can be either extended or merged as soon as possible.

### DIFF
--- a/doc/source/reporting.rst
+++ b/doc/source/reporting.rst
@@ -126,6 +126,8 @@ Others:
      >>> k1 == 'foo:a-b-c'
      True
 
+     Notice that a Key has the same hash, and compares equal (`==`) to its ``str()``.
+
    - in a partial sum over one dimension, e.g. summed along c with dimensions
      a and b:
 
@@ -138,22 +140,22 @@ Others:
      >>> k1.drop('a', 'c') == k2.drop('a') == 'foo:b'
      True
 
-   Some notes:
+   .. note::
+        Some remarks:
 
-   - A Key has the same hash, and compares equal (NOT CLEAR THIS SENTENCE) to its ``str()``.
-     ``repr(key)`` prints the Key in angle brackets ('<>') to signify it is a Key object.
+        - ``repr(key)`` prints the Key in angle brackets ('<>') to signify it is a Key object.
 
-     >>> repr(k1)
-     <foo:a-b-c>
+          >>> repr(k1)
+          <foo:a-b-c>
 
-   - Keys are *immutable*: the properties :attr:`name`, :attr:`dims`, and :attr:`tag` are read-only, and the methods :meth:`append`, :meth:`drop`, and :meth:`add_tag` return *new* Key objects.
+        - Keys are *immutable*: the properties :attr:`name`, :attr:`dims`, and :attr:`tag` are read-only, and the methods :meth:`append`, :meth:`drop`, and :meth:`add_tag` return *new* Key objects.
 
-   - Keys may be generated concisely by defining a convenience method:
+        - Keys may be generated concisely by defining a convenience method:
 
-     >>> def foo(dims):
-     >>>     return Key('foo', dims.split())
-     >>> foo('a b c')
-     foo:a-b-c
+          >>> def foo(dims):
+          >>>     return Key('foo', dims.split())
+          >>> foo('a b c')
+          foo:a-b-c
 
 
 Computations
--- a/doc/source/reporting.rst
+++ b/doc/source/reporting.rst
@@ -98,9 +98,9 @@ Others:
         foo
 
       .. note::
-         Use care when adding literal :class:`str` values (2); these may
-         conflict with keys that identify the results of other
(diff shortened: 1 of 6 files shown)


doc/source/reporting.rst
```diff
@@ -138,7 +140,7 @@ Others:
 
    Some notes:
 
-   - A Key has the same hash, and compares equal to its ``str()``.
+   - A Key has the same hash, and compares equal (NOT CLEAR THIS SENTENCE) to its ``str()``.
```
francescolovat: I don't understand what you meant in this sentence, @khaeru . I left that comment there, pending to be removed in the next commit.

khaeru:
@francescolovat thank you for the attention to detail here! Super appreciated.

doc/source/reporting.rst
khaeru: "compares equal" means using the `==` operator in Python. Sometimes `==` is picky, e.g.:
```
>>> 1 == '1'
False
```
This happens because `1` is an int, and `'1'` is a string.

If you have a Key object on one side of `==`, and a string object on the other side, it will return True. That's showed further up on lines 125–127:
```
>>> k1 = Key('foo', ['a', 'b', 'c'])
>>> k1 == 'foo:a-b-c'
True
```
Here, `k1` is the Key object, and `'foo:a-b-c'` is a string object.

This functionality is provided by `Key.__eq__`: https://github.com/iiasa/ixmp/blob/480d0d81e74e303dd2e3ef1aed3007abbfc3618a/ixmp/reporting/key.py#L93-L95

With this explanation, I'd welcome your suggestion on how to rephrase to make the documentation clearer.

ixmp/reporting/__init__.py
```diff
@@ -480,12 +480,12 @@ def aggregate(self, qty, tag, dims_or_groups, weights=None, keep=True,
         return self.add(key, comp, strict=True, index=True, sums=sums)
 
     def disaggregate(self, qty, new_dim, method='shares', args=[]):
-        """Add a computation that disaggregates *var* using *method*.
+        """Add a computation that disaggregates *qty* using *method*.
```
francescolovat: please @khaeru confirm that the **qty** argument of this method could actually be any quantity (params, variables, etc.) or it has to be just a variable (you wrote **var** probably due to some reason).

doc/source/reporting.rst
```diff
@@ -82,17 +82,19 @@ Others:
...
         >>>     return 'foo'
-        >>> r.add('my report', (my_report, 'scenario'))
-        >>> r.finalize(scenario)
-        >>> r.get('my report')
+        >>> rep.add('my report', (my_report, 'scenario'))
+        >>> rep.finalize(scenario)
+        >>> rep.get('my report')
         foo
 
       .. note::
```
francescolovat: please @khaeru confirm what you meant here with values (2). Was (2) a reference? 

Thanks.

doc/source/reporting.rst
francescolovat: Thank you 😃

doc/source/reporting.rst
khaeru: @francescolovat good question.

Basically, the `.. automethod: add` prints the docstring in the Python code first: https://github.com/iiasa/ixmp/blob/480d0d81e74e303dd2e3ef1aed3007abbfc3618a/ixmp/reporting/__init__.py#L208-L227 …followed by the additional documentation in this .rst file. So if you look at this rendered HTML, line 101 appears after the list where it says “2. any other literal value or constant.”

To explain further:
```
>>> rep.add('y', 'x')
>>> rep.get('y')  # returns the string 'x'
'x'
>>> rep.add('x', 42)  # now 'x' above will be interpreted as a reference to this new node 'x'
>>> rep.get('y')
42
```

This is why [`partial()`](https://docs.python.org/3/library/functools.html#functools.partial) is used at many places in the code. Also, I am summarizing [this part of the dask documentation](https://docs.dask.org/en/latest/spec.html).

Again, your suggestion on how to make this more clear very much appreciated.

ixmp/reporting/__init__.py
khaeru: You're right, it's any quantity :+1:

doc/source/reporting.rst
francescolovat: image

I was asking because unfortunately line 101 does not appear right after “_2. any other literal value or constant._”, it is indeed located far below where the reader loses the perception that that `(2)` actually refers to the second point of the description of the _computation_ argument. I've committed a suggestion to make the note more clear. I hope it will be OK, @khaeru, after that, I think this PR is ready to merge 😃 .

khaeru: I like the new version! Thanks.

PR #232, opened 2019-12-05
Fix exposing whole config file to log output

zikolach:
- fix returning db url as second element of tuple
- check for folder (or empty string) specified as platform dbprops argument
- add test for jdbc backend init output

Checklist
- [x] Tests added.
- [x] Release notes updated.


peterkolp:
@zikolach 
fttb ixmp@5c146f01af491cdd8373aa5a775d5ebade260a37 is deployed to ene-docker.iiasa.ac.at/ixmp-importer:0.2.1
i.e. properties are not exposed.

khaeru:
@zikolach thanks for this change! Please let me know when you've completed the PR checklist and I will review.

zikolach:
@khaeru I added line to release notes. Not sure if we need any kind of documentation as it seems to be bug fix (correct me if I am wrong). Please review/merge.

khaeru:
> Not sure if we need any kind of documentation as it seems to be bug fix

I agree, just wanted to see some kind of explicit statement to that effect in the thread, rather than an unchecked box :slightly_smiling_face:

tests/backend/test_jdbc.py
```diff
@@ -37,6 +37,14 @@ def test_close(test_mp, caplog):
         'Database connection could not be closed or was already closed'
 
 
+def test_connect_message(caplog, request):
+    sample_props = os.path.join(request.fspath.dirname,
+                                '../data/testdb/test.properties.sample')
```
khaeru: Please use the test_data_path fixture: https://github.com/iiasa/ixmp/blob/077387fbabb2536927b723b4e2004810259a3af5/tests/conftest.py#L42-L45
as elsewhere in the test suite. Something like:
```
def test_connect_message(caplog, test_data_path):
    sample_props = test_data_path / 'testdb' / 'test.properties.sample'
    # etc.
```

zikolach: thanks for suggestion - always struggling to reuse fixtures

PR #237, opened 2019-12-16
Fix reading timeseries/layers subannual value

zikolach:
- update ixmp.jar (iiasa/ixmp_source#251, diff https://github.com/iiasa/ixmp_source/compare/78b211703d85d47271018de721c6dbc9b45dc01a...290e790c272c30d7ec3572325ba33154f78f948e)
- fix reading time span DB values mapping
- reduce log level to debug for reading app configuration


khaeru:
Is there a change in the user behaviour here?

zikolach:
@khaeru my comment from slack discussion.

> nikolay.kushin 3:09 PM
> I don't think so. If user's behaviour was to ignore time attribute (sub-annual time reference) assuming timeseries data is always bound to years - then I don't think anybody will have any difference. If users explicitly specified 'year' passing it as parameter time to add_timeseries then it might be difference.

khaeru:
@zikolach in bc42f2d notice that I adjusted the base.Backend.set_meta docstring to actually describe the behaviour as implemented by JDBCBackend. It is very important to keep these in sync. This will also require a test to keep coverage high.

There is 1 test failure due to pint 0.10 being released. I will fix that separately then restart tests here.

ixmp/backend/jdbc.py
```diff
@@ -654,7 +654,14 @@ def unwrap(v):
                 for entry in self.jindex[s].getMeta().entrySet()}
 
     def set_meta(self, s, name, value):
-        self.jindex[s].setMeta(name, value)
+        _type = type(value)
+        try:
+            _type = {int: 'Num', float: 'Num', str: 'Str', bool: 'Bool'}[_type]
+            method_name = 'setMeta' + _type
+        except KeyError:
+            raise TypeError('Cannot store metadata of type {}'.format(_type))
```
khaeru: Please add, at least, a test that covers this line.

khaeru:
CI checks all pass as of 4bbce10.

ixmp/backend/jdbc.py
khaeru: > Please add, at least, a test that covers this line.

I did this in 10713f8.

PR #240, opened 2019-12-18
Add 'ixmp list' CLI command

### LINKED ISSUE #202: Add 'ixmp list' CLI command to print the list of model/scenario strings on a platform

khaeru:
Closes #202.

## How to review

- Run the command against a platform/backend with many Scenarios, e.g. `$ ixmp --platform GP3 list`.
- Run `$ ixmp list --help`, view and try the different options.
- Is there any other functionality that would be useful?

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -18,6 +18,7 @@ Configuration for ixmp and its storage backends has been streamlined.
 
 ## All changes
 
+- [#240](https://github.com/iiasa/ixmp/pull/240): Add ``ixmp list`` command-line tool.
 - [#225](https://github.com/iiasa/ixmp/pull/225): Ensure filters are always converted to string.
 - [#189](https://github.com/iiasa/ixmp/pull/189): Identify and load Scenarios using URLs.
 - [#182](https://github.com/iiasa/ixmp/pull/182),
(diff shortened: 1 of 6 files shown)


francescolovat:
@khaeru thank you for your contribution. I think this is really useful.

I've check the `list` CLI and it works nicely in my PC as well. I've also tested the options in my local database.

I don't completely get what the purpose of the option `--default-only`, is it just showing scenarios with just 1 version? (I mean created with the argument `version='new'` and never ran again?)

I think this PR is **ready to merge**.

khaeru:
Thanks for the review!

> I don't completely get what the purpose of the option `--default-only`, is it just showing scenarios with just 1 version? (I mean created with the argument `version='new'` and never ran again?)

Good question…this is just mirroring the [‘default’ parameter to `Platform.scenario_list()`](https://message.iiasa.ac.at/projects/ixmp/en/master/api-python.html#ixmp.Platform.scenario_list):

> default (bool, optional) – Return only the default version of each TimeSeries/Scenario (see TimeSeries.set_as_default()). Any (model, scenario) without a default version is omitted. If False, return all versions.

That behaviour is a little counter-intuitive, in my opinion…even the idea of having a "default" version of a scenario is kind of an imperfect copy of what git does with chains of commits and branches and the HEAD pointer.

But anyway, this PR doesn't try to change that behaviour (although we could change it later). It just provides a CLI flag to give a True or False value for this parameter.

PR #242, opened 2019-12-19
added --editable to install docu

francescolovat:
Closely related with PR iiasa/message_ix#291

Added `--editable` also in the point `6.` of the installation step from source. Users who'd run that command after installing ixmp using `--editable` in step `5.` would deprecate that flag and therefore not being able to run the tests.

- [x] Documentation added.


khaeru:
Thanks @francescolovat!

PR #243, opened 2020-01-07
Expose timeseries data export

zikolach:
This PR is supposed to be used to draft an API specification and discuss (possible) additional requirements to bulk timeseries data export (filtering/formats etc)

- add platform method to export timeseries data directly to file
- add test


zikolach:
@khaeru thanks for suggestion - improved a bit docstrings and added PR to release notes

zikolach:
@khaeru all you wrote down sound good. I have only one comment - you described the behavior in context of single scenario. The method implemented in this PR covers the case when user needs data for multiple scenarios (potentially whole database content = db snapshot).

khaeru:
> you described the behavior in context of single scenario

In fact, I wrote:

> …ixmp treats two types of data, namely:
> 
> 1. time-series data (non-model; often spanning **many scenarios**) vs.
> 2. model items (parameters, variables, equations, and sets, specific to **one scenario**).

It is conceivable that one would want to dump out, say, the `emissions_factor` parameter values across multiple MESSAGE (model, scenario) combinations. That would be perfectly sensible. The fact that we can't currently do so is just an artifact of how the code was developed.

As I said, I think this distinction (A, time-series data only across multiple scenarios **versus** B, parameters etc. only in a single scenario) is meaningless from a conceptual standpoint and therefore to users. Personally, it created a lot of confusion when starting to work with ixmp. So I would like to erase it as much as possible.

So this is why I propose that, instead of having two methods to write each of A vs. B, then another two methods to read A vs. B, we have a single Backend method for each of reading and writing. Then the JDBCBackend can call different Java functions depending on the arguments; but other Backend implementations are not forced to reproduce the limitations of the Java code.

ixmp/backend/base.py
```diff
@@ -192,6 +193,83 @@ def set_unit(self, name, comment):
...
+        """OPTIONAL: Write Platform, TimeSeries, or Scenario data to file.
+
+        A backend **may** implement write_file for one or more combinations of
+        the `path` and `item_type` methods. For all other combinations, it
+        **must** raise :class:`NotImplementedError`.
+
+        Parameters
+        ----------
+        ts : TimeSeries or None
+            Single TimeSeries object to read.
```
zikolach: write, not read

ixmp/backend/base.py
```diff
@@ -192,6 +193,83 @@ def set_unit(self, name, comment):
...
+        path : os.PathLike
+            File for output. The filename suffix determines the output format.
+        item_type : ItemType
+            Type(s) of items to write.
+        filters : dict of dict of str
```
zikolach: should it be either ts passed as parameter, or filters

khaeru:
Yes, I agree it's awkward.

TimeSeries (therefore also Scenario) has a shorthand/helper method here:
https://github.com/iiasa/ixmp/blob/c2e892911280c18030a48fc75997521d4701056c/ixmp/core.py#L325-L327

This helps calling the many Backend methods where the first argument is a TimeSeries/Scenario instance. Some of the other Backend methods are Platform-wide, e.g. for units—these ones don't take an object instance as the first argument.

So the proposed code tries to mimic the first style but allow `ts=None`. The alternate would be to mimic the second style, so delete the first argument. That would require the following changes:

- Look for `filters=dict(scenario=[ts])` where `ts` is a TimeSeries/Scenario object, and then use that object for the rest of the I/O code. That means `filters` is a dict of dict of strings *except* for that one possibility…also a little confusing.
- Change the calls here: https://github.com/iiasa/ixmp/blob/52451370958bebc9bab059c7caaa7436c60d8811/ixmp/model/gams.py#L130-L145
  Currently this makes use of the _backend() shorthand method on the Scenario object.

I'll try this now and we'll see if it looks less awkward.

RELEASE_NOTES.rst
```diff
@@ -4,6 +4,7 @@ Next release
 All changes
 -----------
 
+- `#243 <https://github.com/iiasa/ixmp/pull/243>`_: Add bulk timeseries data export
```
khaeru: Link to the docs for the new method, and check that the link works in the docs build. End with a period.
```
Add :meth:`.export_timeseries_data` to write data for multiple TimeSeries objects to CSV.
```

ixmp/core.py
```diff
@@ -164,6 +164,48 @@ def scenario_list(self, default=True, model=None, scen=None):
         return pd.DataFrame(self._backend.get_scenarios(default, model, scen),
                             columns=FIELDS['get_scenarios'])
 
+    def export_timeseries_data(self, path, default=True, model=None,
+                               scenario=None, variables=None):
+        """Bulk timeseries export to CSV file.
```
khaeru: - I would avoid introducing the word "bulk", which we don't define and don't use anywhere else.
- Explain that only data added with `add_timeseries` (link to the method) is exported.

ixmp/core.py
```diff
@@ -164,6 +164,48 @@ def scenario_list(self, default=True, model=None, scen=None):
...
+
+        Parameters
+        ----------
+        file : os.PathLike
+            File name to export data to; must have the suffix '.csv'.
+
+            Result file will contain the following columns:
```
khaeru: I think we need to move this to a new section of the documentation, which will also contain descriptions of the GDX and Excel file formats.

That can either be in api-backend.rst (named like “File formats”) or in api-python.rst—not sure where is best.

Maybe okay to leave this where it is for now and address that when we bring in the Excel I/O code from message_ix.

ixmp/core.py
```diff
@@ -164,6 +164,48 @@ def scenario_list(self, default=True, model=None, scen=None):
...
+            - variable
+            - unit
+            - region
+            - meta
+            - time
+            - year
+            - value
+        default : bool, optional
+            :obj:`True` to include only TimeSeries versions marked as default.
+        model : str or None, optional
```
khaeru: Is this correct? Can't it be a list of str? cf. https://github.com/iiasa/ixmp/blob/fa5139b7659d416abb3f5fe79c5f09139d02779b/ixmp/backend/jdbc.py#L385

Similar with the `scenario` arg. `variables` should be "list of str"

zikolach: > Is this correct? Can't it be a list of str? cf.

As far as I see it is correct. The method signature mimics `get_scenarios` which can also take model/scenario names as parameters expecting string or None to be passed.

ixmp/core.py
```diff
@@ -164,6 +164,51 @@ def scenario_list(self, default=True, model=None, scen=None):
...
+        -------
+        None
+        """
+        filters = {
+            'model': model,
+            'scenario': scenario,
+            'variable': as_str_list(variables) or [],
+        }
+
+        self._backend.write_file(path, ItemType.TS, filters, default=default)
```
zikolach: @khaeru is there any particular reason not putting `default` (or `is_default` if better) to filters?

ixmp/core.py
zikolach: Fine with moving later.

zikolach:
> TimeSeries (therefore also Scenario) has a shorthand/helper method here

@khaeru can you please explain/comment why it is called a "convenience" method?

```
self._backend('set_data', r, v, data.astype(float).dropna(), u, meta)
```
vs
```
self._backend.set_data(r, v, data.astype(float).dropna(), u, meta)
```
Is it following some recommendations/guidelines for Python?
In particular it breaks code navigation in Pycharm/IntelliJ IDEA. Which benefits it brings in return?

ixmp/core.py
khaeru: Okay! Please open an issue for it so we don't forget.

ixmp/core.py
khaeru: I see—so the functionality provided for `model`/`scenario` and for `variable` are not the same?
1. For `model`, you can optionally provide a single model name to restrict to only that model.
2. For `scenario`, the same.
3. For `variable`, you can provide a list of one or more names?

If that's the case, the docstring should be like:
```
model: str, optional
    Only return data for this model name.
scenario: str, optional
    Only return data for this scenario name.
variable: list of str, optional
    Only return data for variable name(s) in this list.
```
I also think `variable` is better than `variables` as an arg name.

ixmp/core.py
khaeru: Elsewhere the code, `filters` are always a `dict` where the keys are the dimensions and the values are `list of str` with allowable values. Only the values appearing in the filters are returned.

You could think of `default` as an attribute that has a True or False value. But elsewhere in the code (e.g. Platform.scenarios_list), a `default=False` argument does *not* mean "_only_ scenarios that have default=False"; instead, it means "all scenarios, regardless of whether default=False or default=True".

I personally think that is confusing, but it's extra confusing to do it two different ways in two places. So that's why, for now, I would keep these separate.

khaeru:
> can you please explain/comment why it is called a "convenience" method?

Many of the Backend methods take a TimeSeries or Scenario as the first argument. The second example:
```
self._backend.set_data(r, v, data.astype(float).dropna(), u, meta)
```
…is not correct; it would need to be:
```
self.platform._backend.set_data(self, r, v, data.astype(float).dropna(), u, meta)
```
It's convenient to not have to write this everywhere. It's also easy to get it wrong, in part because using 'self' twice _looks_ un-Pythonic (even though it's correct). I'm sorry it confuses your editor!

ixmp/core.py
zikolach: > I also think variable is better than variables as an arg name.

Didn't get the idea why it is better, but if you think so I'll change it

khaeru: For example, when we handle a `filters` dictionary, if the dimension is named 'technology' or 'year', then 'technology'/'year' is what one uses as a key in that dict…even if it maps to a list of technologies or years (plural). `variable` is, essentially, another dimension, so IMO should be treated the same way.

ixmp/backend/base.py
```diff
@@ -192,6 +193,95 @@ def set_unit(self, name, comment):
...
+            Suffix Format
+            ====== ===
+            .csv   Comma-separated values
+            .gdx   GAMS data exchange
+            .xlsx  Microsoft Office Open XML spreadsheet
+            ====== ===
+
+        item_type : ItemType
+            Type(s) of items to read.
+        filters : dict of dict of str
```
zikolach: @khaeru is it `dict of dict of str`?

what do you think about moving `filters` for both methods to `kwargs`? (as it seems to be optional parameter in some cases)

khaeru: `dict (str -> list of str)`—sorry, that's my mistake.

khaeru: > moving `filters` for both methods to `kwargs`

Yes, that's a good idea. Nice catch! I think `item_type` needs to say as it very much changes how the backend handles the function call.

zikolach:
> …is not correct; it would need to be:

You are right - call from TimeSeries/Scenario will look shorter/less repetitions.

> I'm sorry it confuses your editor!

I hope that's not sarcastic reaction ;) as navigating the code recently is getting more difficult. Can share your experiences/best practices, if you don't have such problems.

PR #260, opened 2020-01-29
Include ixmp.Platform.units() in docs

francescolovat:
Wondering how to obtain a full list of units from _MESSAGEix_ I found that a _**docstring**_ for the method `.units()` of `ixmp.Platform` was missing and therefore not reported in the documentation.


francescolovat:
Hi @khaeru , 

I think this PR is ready to be merged. Please approve if you think that's the case.

khaeru:
```
$ git checkout docu-ixmp.Platfrom.units()
bash: syntax error near unexpected token `('
```

In the future, please just use lower-case characters plus '-' and/or '/' in branch names!

PR #262, opened 2020-01-30
Give more helpful messages on "ixmp list" user error

francescolovat:
This PR helps `ixmp` **CLI** to create more intuitive error messages when user fails to insert the _commands_ or _options_ as it is required for it work.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,8 +1,8 @@
 from pathlib import Path
+from pandas.testing import assert_frame_equal
 
 import ixmp
 import pandas as pd
-import pandas.testing as pdt
 
 
 def test_main(ixmp_cli, test_mp, tmp_path):
@@ -147,4 +147,4 @@ def test_import(ixmp_cli, test_mp, test_data_path):
         'model': ['canning problem'],
         'scenario': ['standard'],
     })
-    pdt.assert_frame_equal(exp, scen.timeseries())
+    assert_frame_equal(exp, scen.timeseries())
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 from pandas.testing import assert_frame_equal
+from click.exceptions import UsageError
 
 import ixmp
 import pandas as pd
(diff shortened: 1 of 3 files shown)


francescolovat:
Hi @khaeru, help here. An error message is properly shown now when no options are given before the command list.

However, not the test_main fails:
```

ixmp_cli = <ixmp.testing.ixmp_cli.<locals>.Runner object at 0x0000019AFC429D30>
test_mp = <ixmp.core.Platform object at 0x0000019AFC678390>
tmp_path = WindowsPath('C:/Users/lovat/AppData/Local/Temp/pytest-of-lovat/pytest-38/test_main0/temp.properties')

    def test_main(ixmp_cli, test_mp, tmp_path):
        # Name of a temporary file that doesn't exist
        tmp_path /= 'temp.properties'

        # Giving --dbprops and a nonexistent file is an invalid argument
        cmd = [
            '--platform', 'pname',
            '--dbprops', str(tmp_path),
            'platform', 'list',  # Doesn't get executed; fails in cli.main()
        ]
        r = ixmp_cli.invoke(cmd)
        assert r.exit_code == 2  # click retcode for bad option usage

        # Create the file
        tmp_path.write_text('')

        # Giving both --platform and --dbprops is bad option usage
        r = ixmp_cli.invoke(cmd)
>       assert r.exit_code == 1
E       assert 2 == 1
E        +  where 2 = <Result SystemExit(2)>.exit_code

tests\test_cli.py:26: AssertionError
```

It depends on the behavior of the following function. Should I just change assert to ==2 ?

https://github.com/francescolovat/ixmp/blob/f9d398670516e06f5d2e3510256ff4ef05b78d10/ixmp/testing.py#L52-L59

khaeru:
In order to understand what the test is doing, look at the click documentation about:
- testing: https://click.palletsprojects.com/en/7.x/testing/#basic-testing

  > The return value is a Result object, which has the captured output data, exit code, and optional exception attached.

- exceptions: https://click.palletsprojects.com/en/7.x/exceptions/

  > If an ClickException is raised, invoke the ClickException.show() method on it to display it and then exit the program with ClickException.exit_code.

  then, lower:

  > Which Exceptions Exist?
  > …
  > - UsageError to inform the user that something went wrong.

Browsing over to the click source code on GitHub: https://github.com/pallets/click/
…one can search for "UsageError", finding it in click/exceptions.py:
https://github.com/pallets/click/blob/8df9a6b2847b23de5c65dcb16f715a7691c60743/click/exceptions.py#L43-L51

From this (also the comment `# click retcode for bad option usage` a few lines above) it should be possible to figure out how to adjust the test. Please give it a try!

khaeru:
PS. even better might be to improve the tests with something like `assert r.exit_code == BadOptionUsage.exit_code`. This avoids having to manually look up the value and enter it in the test file, and you can see that the exit_code is the one associated with the exception `raise`d in the code.

francescolovat:
Hi @khaeru,

also here. Please approve if you find the task is completed.

tests/test_cli.py
```diff
@@ -15,30 +16,31 @@ def test_main(ixmp_cli, test_mp, tmp_path):
         '--dbprops', str(tmp_path),
         'platform', 'list',  # Doesn't get executed; fails in cli.main()
     ]
-    r = ixmp_cli.invoke(cmd)
-    assert r.exit_code == 2  # click retcode for bad option usage
+    result = ixmp_cli.invoke(cmd)
+    assert result.exit_code == UsageError.exit_code  # click retcode for bad
```
francescolovat: > PS. even better might be to improve the tests with something like `assert r.exit_code == BadOptionUsage.exit_code`. This avoids having to manually look up the value and enter it in the test file, and you can see that the exit_code is the one associated with the exception `raise`d in the code.

I guess this is what you meant

khaeru: Exactly so!

khaeru:
For some reason the list of commits here is showing some unrelated commits from `master`. I'm not sure why that is—will try a rebase to clean it up.

khaeru:
Okay, that worked. I will make the same fix to the branch for #260, and wait for CI to pass.

PR #264, opened 2020-02-04
Extend API to support dynamic time slices

zikolach:
[Requirements](https://github.com/iiasa/ene/wiki/Requirements-sub-annual-timeseries)

Changelist:
- new method to add time slices to DB
- new method to get list of time slices from DB
- input/output dataframes containing timeseries data now may include **subannual** column (containing name of time slice)
- relevant changes in ixmp.jar (source diff https://github.com/iiasa/ixmp_source/compare/0cd6eb355af454e3c7d214d0e5612ea455aff9c0...10742197121eaf83c6f222f969d7e2842d0bbdde)
- update tests suite

Actions:
- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.
- [x] Update this PR with ixmp.jar built from ixmp_source master

Related PR https://github.com/iiasa/ixmp_source/pull/264

PR #269, opened 2020-02-28
add docstrings related to the `timeslices` feature

danielhuppmann:
## Description

This PR adds docstrings to the branch in #264 as requested by @zikolach [here](https://github.com/iiasa/ixmp/pull/264#issuecomment-583406552).

## How to review

Please read the added docstrings and comment whether they clarify the intention of this feature.
Also, I'm not sure whether there aren't "better" places to put this info (e.g., the backend?) or whether this should be duplicated somewhere...

### DIFF
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -284,6 +284,10 @@ class Platform:
         category (e.g., all months) from the pandas.DataFrame returned by this
         function or to aggregate subannual data to full-year results.
 
+        A timeslice is related to the index set 'time' in a message_ix.Scenario
+        to indicate a subannual temporal dimension. Alas, timeslices and set
+        elements of time have to be initialized/defined independently.
+
         See :meth:`add_timeslice` to initialize additional timeslices in the
         Platform instance.
 
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -271,7 +271,21 @@ class Platform:
         self._backend.set_node(region, synonym=mapped_to)
 
     def timeslices(self):
-        """Return all sub-annual time slices.
+        """Return all subannual time slices defined in this Platform instance.
+
+        Timeslices are a way to represent subannual temporal resolution in
+        timeseries data. A timeslice consists of a **name** (e.g., 'January',
+        'summer'), a **category** (e.g., 'months', 'seasons'), and a
+        **duration** given relative to a full year.
+
+        The category and duration do not have any functional relevance within
+        the ixmp framework, but they may be useful for pre- or post-processing.
+        For example, they can be used to filter all timeslices of a certain
+        category (e.g., all months) from the pandas.DataFrame returned by this
+        function or to aggregate subannual data to full-year results.
+
+        See :method:`add_timeslice` to initialize additional timeslices in the
+        Platform instance.
 
         Returns
         -------
@@ -281,14 +295,16 @@ class Platform:
                             columns=FIELDS['get_timeslices'])
 
     def add_timeslice(self, name, category, duration):
-        """Define a sub-annual time slice including a category and duration.
+        """Define a subannual time slice including a category and duration.
+
+        See :method:`timeslices` for a detailed description of timeslices.
 
         Parameters
         ----------
         name : str
             Unique name of the time slice.
         category : str
-            Time slice category (e.g. COMMON, MONTHS etc).
+            Time slice category (e.g. 'common', 'month', etc).
         duration : float
             Duration of time slice as fraction of year.
         """
@@ -448,6 +464,13 @@ class TimeSeries:
 
             - `year` and `value`—long, or 'tabular', format.
             - one or more specific years—wide, or 'IAMC' format.
+
+            To support subannual temporal resolution of timeseries data, a
+            column `subannual` is optional in `df`. The entries in this column
+            must have been defined in the Platform instance using
+            :method:`add_timeslice` beforehand. If no column `subannual` is
+            included in `df`, the data is assumed to contain yearly values.
+            See :method:`timeslices` for a detailed description of this feature.
 
         meta : bool, optional
             If :obj:`True`, store `df` as metadata. Metadata is treated
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -274,7 +274,7 @@ class Platform:
         """Return all subannual timeslices defined in this Platform instance.
 
         Timeslices are a way to represent subannual temporal resolution in
-        timeseries data. A timeslice consists of a **name** (e.g., 'January',
+        timeseries data. A timeslice consists of a **name** (e.g., 'january',
         'summer'), a **category** (e.g., 'months', 'seasons'), and a
         **duration** given relative to a full year.
 
--- a/ixmp/core.py
+++ b/ixmp/core.py
@@ -271,7 +271,7 @@ class Platform:
         self._backend.set_node(region, synonym=mapped_to)
 
     def timeslices(self):
-        """Return all subannual time slices defined in this Platform instance.
+        """Return all subannual timeslices defined in this Platform instance.
 
         Timeslices are a way to represent subannual temporal resolution in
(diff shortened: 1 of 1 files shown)


ixmp/core.py
```diff
@@ -287,7 +287,21 @@ def add_region_synonym(self, region, mapped_to):
         self._backend.set_node(region, synonym=mapped_to)
 
     def timeslices(self):
-        """Return all sub-annual time slices.
+        """Return all subannual timeslices defined in this Platform instance.
+
+        Timeslices are a way to represent subannual temporal resolution in
+        timeseries data. A timeslice consists of a **name** (e.g., 'January',
```
zikolach: nit: we should maybe be consistent about upper/lower/capital case referring timeslices, otherwise wi end up having SUMMER, summer, Summer vales as timeslices

ixmp/core.py
```diff
@@ -465,6 +481,13 @@ def add_timeseries(self, df, meta=False):
             - `year` and `value`—long, or 'tabular', format.
             - one or more specific years—wide, or 'IAMC' format.
 
+            To support subannual temporal resolution of timeseries data, a
+            column `subannual` is optional in `df`. The entries in this column
```
zikolach: It may be a confusion as somewhere it called `time`, not `subannual`

ixmp/core.py
danielhuppmann: good point - I'd go for lower-case wherever possible. updated.

ixmp/core.py
danielhuppmann: where is it called `time` in the public-facing part?

zikolach:
@danielhuppmann I rebased source PR as Paul suggested in comments. Please rebase your commits onto it.

danielhuppmann:
> @danielhuppmann I rebased source PR as Paul suggested in comments. Please rebase your commits onto it.

done

ixmp/core.py
```diff
@@ -271,7 +271,21 @@ def add_region_synonym(self, region, mapped_to):
...
+        the ixmp framework, but they may be useful for pre- or post-processing.
+        For example, they can be used to filter all timeslices of a certain
+        category (e.g., all months) from the pandas.DataFrame returned by this
+        function or to aggregate subannual data to full-year results.
+
+        See :method:`add_timeslice` to initialize additional timeslices in the
+        Platform instance.
 
         Returns
         -------
```
khaeru: - What are the contents of the returned object?

ixmp/core.py
```diff
@@ -271,7 +271,21 @@ def add_region_synonym(self, region, mapped_to):
...
     def timeslices(self):
-        """Return all sub-annual time slices.
+        """Return all subannual timeslices defined in this Platform instance.
+
+        Timeslices are a way to represent subannual temporal resolution in
+        timeseries data. A timeslice consists of a **name** (e.g., 'january',
+        'summer'), a **category** (e.g., 'months', 'seasons'), and a
+        **duration** given relative to a full year.
+
+        The category and duration do not have any functional relevance within
```
khaeru: Clear _negative_ statements like this are super helpful :+1:

Along the same lines, do the durations within a category need to sum to 1? Or does the code not care?

ixmp/core.py
```diff
@@ -271,7 +271,21 @@ def add_region_synonym(self, region, mapped_to):
...
+        the ixmp framework, but they may be useful for pre- or post-processing.
+        For example, they can be used to filter all timeslices of a certain
+        category (e.g., all months) from the pandas.DataFrame returned by this
+        function or to aggregate subannual data to full-year results.
+
+        See :method:`add_timeslice` to initialize additional timeslices in the
```
khaeru: I believe the ReST role is `:meth:`. Does this link compile correctly?

ixmp/core.py
```diff
@@ -281,16 +295,18 @@ def timeslices(self):
                             columns=FIELDS['get_timeslices'])
 
     def add_timeslice(self, name, category, duration):
-        """Define a sub-annual time slice including a category and duration.
+        """Define a subannual timeslice including a category and duration.
+
+        See :method:`timeslices` for a detailed description of timeslices.
```
khaeru: What happens if the name is already defined?

As with units, suppose I accidentally define a timeslice with a typo. Can I delete it?

ixmp/core.py
danielhuppmann: Code doesn't care...

ixmp/core.py
danielhuppmann: > What happens if the name is already defined?

@zikolach, we could add something on the Python side like:
```
if name in self.timeslices.name:
    logger().info('Timeslice {} already defined, skipped this command'.format(name))
    return
... current function ...
```

> As with units, suppose I accidentally define a timeslice with a typo. Can I delete it?

No API yet, you'd have to do it directly via the underlying database. This is similar to the regions API. We'll add it at a later point when the usage becomes clearer. And before implementation of a delete feature, we'll need to discuss more: who can delete, how to treat existing data upon deletion, etc....

ixmp/core.py
danielhuppmann: done

ixmp/core.py
danielhuppmann: added a line

danielhuppmann:
> Perhaps this needs to be reviewed as part of the parent PR, but the column name does appear to be 'time', at least for geodata. Please also consider consistency with ixmp.Scenario / message_ix model data, which uses the name 'time' as an index set. (Also, explain to users that adding a timeslice/using it for time-series is completely distinct from dimensions of the same name in message_ix.)

We have an unresolvable naming conflict between MESSAGE (where `time` is the set of subannual time periods) and the IAMC format, where `time` is used by **openscm** (and therefore pyam) for continuous-time format.

I'll leave it to @zikolach to decide whether to refactor the backend and database schema to call the column `TIMESLICE` or `SUBANNUAL` for consistency.

(I added a sentence in the docstring to highlight the relation between timeslices and time).

> Further thoughts: 'subannual' would further muddy already troubled waters.

@khaeru, if you have a better, actionable suggestion on how to name this column (instead of `subannual`, I'm all ears... (see above, `time` is not an option).

> Although we use the term 'year' frequently, sometimes this is an annual value at one point in time, and sometimes it is a total/average/etc. over a time _period_, whereof the 'year' is the start, middle, or end, and whose duration is implicitly defined by the gap to the previous/subsequent 'year'.

I doubt that we'll be able to resolve the inherent ambiguity of modelling in this PR. We're trying to implement something that works for given use cases with the time & resources available.
 
> But nothing in the ixmp/message_ix stack, AFAIK, prevents one putting 1..365 in the 'year' column and saying that you have an energy systems model that covers one year, in which the duration of each period is 1 day. Then 'time' would not be sub-annual, but sub-daily, etc.

Yes, one can use the format in an unintended way.

The intended way to store one-year hourly data this in the extended IAMC-format would be:
 - use a unique value for `year`
 - use timeslices like `jan-01-01 01:00, jan-01-01 02:00, jan-01-01 03:00,`, etc.

> Anyway, I'm clicking "approve" in case you want to merge this child PR and continue in the parent one, but I hope these things can be clarified by the final documentation.

Thanks, over to @zikolach for merging (after unit tests pass)!

ixmp/core.py
zikolach: I guess as dataframe column name. Just wanted to make an accent that we may want to rename it (e.g. in parent PR). See test_new_subannual_timeseries_as_iamc test

danielhuppmann: For me, it's important that all public-facing interfaces (both getter and setter) in `ixmp` and the RestAPI use the column name `subannual` and `timeslices` for the allowed values (unless someone comes up with an alternative in the next 48 hours). Whether it makes sense to change that throughout the ixmp Java backend and the database schemes - I leave that decision to you.

The test `test_new_subannual_timeseries_as_iamc` is as I expect it (well, preferably lower-case `summer`).

PR #276, opened 2020-03-06
Change timeslice tests to expected behaviour

zikolach:
**Required:** write a single sentence here describing the changes made by this PR.

## How to review

**Required:** describe specific things that reviewer(s) need to do to, in order
to ensure that the PR does what it intends.

## PR checklist

- [ ] Tests added.
- [ ] Documentation added.
- [ ] Release notes updated.

PR #277, opened 2020-03-09
Add tests and rename time to subannual

zikolach:
Replaces #276 fixing and extending original changes


ixmp/core.py
```diff
@@ -490,10 +502,10 @@ def add_timeseries(self, df, meta=False):
         # Ensure consistent column names
         df = to_iamc_template(df)
 
-        if 'time' not in df.columns:
-            df['time'] = 'Year'
+        if 'subannual' not in df.columns:
```
danielhuppmann: setting the default should happen inside the Java core, in my opinion

ixmp/backend/jdbc.py
```diff
@@ -467,11 +467,14 @@ def get_data(self, ts, region, variable, unit, year):
             'value': float,
         }
 
+        fields_map = {'subannual': 'time'}
```
danielhuppmann: renaming should happen inside the Java core, in my opinion, so that the renaming shouldn't be required in every API (ixmp, restAPI, maybe more in the future)

ixmp/core.py
zikolach: it is done like this for simplicity - java already set default value if passed null value as timeslice name. Feel free to rewrite this method to incorporate your suggested change.

ixmp/backend/jdbc.py
zikolach: it will break web API, possibly Scenario Explorer and server integration scripts and also introduce inconsistency in java code and DB model (if done without db migration)

danielhuppmann: I'm aware that this is a major refactoring, but it is better to break the API now (before there are more active users) rather than maintaining inconsistent APIs going forward. Also, we discussed the requirement that this column be called `subannual` for weeks, so this should not come as a surprise.

ixmp/core.py
zikolach: please have a look on updated implementation

ixmp/backend/jdbc.py
zikolach: this PR contains jar returning subannual as timeseries attribute, therefore rename is removed

danielhuppmann:
took a first glance at the changes, looks good to me - have one more idea for an additional test, will try to implement and PR into this branch tomorrow

zikolach:
> took a first glance at the changes, looks good to me - have one more idea for an additional test, will try to implement and PR into this branch tomorrow

Will it make sense to merge this and target your new PR to original branch?

zikolach:
Merging to original PR #264 to continue

PR #280, opened 2020-03-13
add warning if importing timeseries with unexpected columns

danielhuppmann:
This PR adds a warning when trying to import a timeseries dataframe with unexpected coumns (e.g., `time`, `climate_model`) - this would have helped me last week to understand why my trials didn't work.

The PR also fixes a problem that timeseries-data-columns were categorised by the content (`dtype`) rather than whether the column name is like a year. A test is added to check for correct behaviour.

### DIFF
--- a/ixmp/tests/core/test_timeseries.py
+++ b/ixmp/tests/core/test_timeseries.py
@@ -65,7 +65,8 @@ def expected(df, ts):
 
 def wide(df):
     """Transform *df* from long to wide format."""
-    return df.pivot_table(index=IAMC_IDX, columns='year', values='value') \
+    other_cols = [c for c in df.columns if c not in ['year', 'value'] + IAMC_IDX]
+    return df.pivot_table(index=IAMC_IDX + other_cols, columns='year', values='value') \
         .reset_index() \
         .rename_axis(columns=None)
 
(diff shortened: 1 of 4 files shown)


danielhuppmann:
@zikolach ready for review

ixmp/tests/core/test_timeseries.py
```diff
@@ -174,6 +174,18 @@ def test_add_timeseries(ts, format):
         ts.add_timeseries(DATA[0].drop('unit', axis=1))
 
 
+@pytest.mark.parametrize('format', ['long', 'wide'])
+def test_add_timeseries_with_extra_col(ts, format):
+    _data = DATA[0].copy()
+    _data['climate_model'] = [0, 1]
+    data = _data if format == 'long' else wide(_data)
```
zikolach: not sure if `climate_model` will be in `data` after calling `wide`

ixmp/utils.py
```diff
@@ -147,11 +146,16 @@ def pd_write(df, f, *args, **kwargs):
...
-    """Return the indices of the numeric columns of *df*."""
-    dtypes = df.dtypes
-    return [i for i in dtypes.index
-            if dtypes.loc[i].name.startswith(('float', 'int'))]
+def year_list(x):
+    """Return the elements of x that can be cast to year (int)."""
+    lst = []
+    for i in x:
+        try:
+            int(i)  # this is a year
```
zikolach: not sure but maybe better to check using regex (something like `^\d{4}$`) if we don't want negative or too small/large number

zikolach:
@danielhuppmann I added some changes to the PR - please have a look

ixmp/utils.py
danielhuppmann: in the MESSAGEix westeros tutorial, we use `years = [690, 700, 710, ...]`, so let's not be overly restrictive

ixmp/tests/core/test_timeseries.py
danielhuppmann: added a check to verify expected behaviour

danielhuppmann:
thanks @zikolach for the additions, should be good to go now

Issue #271, opened 2020-03-03
Improving Reporter.describe() function 


awais307:
Currently, if we use Reporter.describe()  it can do one of two things; 
1. `Reporter.describe()` -- print AND return the string 
2. `Reporter.describe(quiet = True)` -- only return the string and don't print it. 

It can be changed that it only returns a string by default and don't print without mentioning `quiet = True `

Reference code to change ([link](https://github.com/iiasa/ixmp/blob/19c19e46367042bc12d8ab2d11d26e9cbd453204/ixmp/reporting/__init__.py#L539-L556))

PR #285, opened 2020-03-20
Reporter.describe modified to not print by default

behnam-zakeri:
Per suggestion by @awais307 via issue #271, this PR slightly modifies the code in `Reporter.describe()` so that the output is not printed by default. 

## How to review
The method can be run without passing any argument to _quiet_, and the results shouldn't be printed.

## PR checklist

- [x] ~Tests added.~ Minor change in behaviour, no test and documentation is needed.

### DIFF
--- a/ixmp/tests/reporting/test_reporting.py
+++ b/ixmp/tests/reporting/test_reporting.py
@@ -491,9 +491,14 @@ def test_reporter_describe(test_mp, test_data_path, capsys):
     - <ixmp.core.Scenario object at {id}>
   - 'config':
     - {{'filters': {{}}}}""".format(id=id_)
-    assert desc1 == r.describe('d:i', quiet=False)
+    assert desc1 == r.describe('d:i')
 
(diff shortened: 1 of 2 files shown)


khaeru:
@awais307 since this is your issue, I'll ask you to work with @behnam-zakeri to ensure the change is as desired.

ixmp/tests/reporting/test_reporting.py
```diff
@@ -491,7 +491,7 @@ def test_reporter_describe(test_mp, test_data_path, capsys):
     - <ixmp.core.Scenario object at {id}>
   - 'config':
     - {{'filters': {{}}}}""".format(id=id_)
-    assert desc1 == r.describe('d:i')
+    assert desc1 == r.describe('d:i', quiet=False)
```
khaeru: Does this actually need to change, or will it still work with the default (quiet=True)?

khaeru:
Excellent! Thanks both.

PR #286, opened 2020-03-20
Transfer Excel data I/O from message_ix

khaeru:
**Replaced by #289.**

PR #289, opened 2020-03-23
Transfer Excel data I/O from message_ix

### LINKED ISSUE #284: Transfer Excel data import from message_ix
As described at https://github.com/iiasa/ixmp/pull/243#pullrequestreview-347406647:

> …the Excel I/O methods (currently hiding in message_ix) can *also* be rolled in (:+1:) as:
> ```
> b.write_file(path='bla.xlsx', ...)
> b.read_file(path='bla.xlsx', what=DataType.MODEL, ...)  # or something like that
> ```

- Necessary for the EOSC project, so that data can be imported into a Scenario to parametrize a specific job/run.
- Supersedes iiasa/message_ix#265; that code should be incorporated here.

khaeru:
This PR streamlines and improves file input/output in the MESSAGE stack:

- `Scenario.read_excel()`, `.to_excel()` are brought to ixmp from message_ix (see iiasa/message_ix#321).
  - Can now be used with any ixmp Scenario.
  - The code is improved in several ways, e.g. it can now distinguish between 1-D index sets and indexed sets; degrades gracefully by giving readable log messages and exceptions.
- `TimeSeries.read_file` is added as a replacement for `ixmp.utils.import_timeseries`.
- New: commands `import` and `export` that allow performing the above operations from the command-line.
- New: documentation that clearly describes the file formats supported (and not supported) for I/O; docstrings for all touched methods.
- New: `TimeSeries.add_timeseries(..., year_lim=(2020, 2080))`, which brings the minimum/maximum year functionality previously only supported by `import_timeseries()` to non-file-based operations.
- See #290 & thanks @zikolach for:
  - ixmp.jar is updated to clean up Java exception handling.
  - {,JDBC}Backend gain a method, `get_log_level()`.

Closes #284.
Replaces #286.

## How to review

- Build and read the documentation.
- Check that any I/O scripts still work.

  One possible breaking change: `import_timeseries()` previously automatically prepended the string 'R11_' to imported region names, unless the name was 'World'. This is specific to MESSAGEix-GLOBIOM (not message_ix), so it is removed.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


doc/source/file-io.rst
```diff
@@ -0,0 +1,78 @@
...
+
+Scenario data can be read from/written to Microsoft Excel files using :meth:`.Scenario.read_excel` and :meth:`.to_excel`, and the CLI commands ``ixmp import scenario FILE`` and ``ixmp export FILE``.
+The files have the following structure:
+
+- One sheet named 'ix_type_mapping' with two columns:
+
+  - 'item': the name of an ixmp item.
+  - 'ix_type': the item's type as a length-3 string: 'set', 'par', 'var', or 'equ'.
+
+- One sheet per item.
```
zikolach: can it handle large items having more 1M elements?

doc/source/file-io.rst
```diff
@@ -0,0 +1,78 @@
...
+  - Sheets for index sets have one column, with a header cell that is the set name.
+  - Sheets for one-dimensional indexed sets have one column, with a header cell that is the index set name.
+  - Sheets for multi-dimensional indexed sets have multiple columns.
+  - Sets with no elements are represented by empty sheets.
+
+- Parameters, variables, and equations:
+
+  - Sheets have zero (for scalar items) or more columns with headers that are the index *names* (not necessarily sets; see below) for those dimensions.
+  - Parameter sheets have 'value' and 'unit' columns.
+  - Variable and equation sheets have 'lvl' and 'mrg' columns.
```
zikolach: any particular reason to shorten names?

doc/source/file-io.rst
```diff
@@ -0,0 +1,78 @@
...
+
+  - Sheets for index sets have one column, with a header cell that is the set name.
+  - Sheets for one-dimensional indexed sets have one column, with a header cell that is the index set name.
+  - Sheets for multi-dimensional indexed sets have multiple columns.
+  - Sets with no elements are represented by empty sheets.
+
+- Parameters, variables, and equations:
+
+  - Sheets have zero (for scalar items) or more columns with headers that are the index *names* (not necessarily sets; see below) for those dimensions.
+  - Parameter sheets have 'value' and 'unit' columns.
```
zikolach: shouldn't unit have same value for all elements of parameter? if yes, maybe have extra sheet - mapping between parameter names and units?

ixmp/backend/base.py
```diff
@@ -742,14 +790,19 @@ def item_get_elements(self, s: Scenario, type, name, filters=None):
             When *type* is 'set' and *name* an index set (not indexed by other
             sets).
         dict
-            When *type* is 'equ', 'par', or 'set' and *name* is scalar (zero-
+            When *type* is 'equ', 'par', or 'var' and *name* is scalar (zero-
```
zikolach: not sure, but maybe `'equ', 'par', 'var', or 'set'`?

ixmp/backend/jdbc.py
```diff
@@ -405,7 +433,15 @@ def get(self, ts, version):
         try:
             jobj = method(*args)
         except java.IxException as e:
-            raise RuntimeError(*e.args) from None
+            # Try to re-raise as a ValueError for bad model or scenario name
+            match = re.search(r"table '([^']*)' from the database", e.args[0])
```
zikolach: maybe better to change exception handling on java side to generate proper error message?

ixmp/backend/jdbc.py
```diff
@@ -603,13 +642,10 @@ def init_item(self, s, type, name, idx_sets, idx_names):
         try:
             func(name, idx_sets, idx_names)
         except jpype.JException as e:
-            e = str(e)
-            if 'This Scenario cannot be edited' in e:
-                raise RuntimeError(e)
-            elif 'already exists' in e:
-                raise ValueError('{!r} already exists'.format(name))
+            if 'already exists' in e.args[0]:
```
zikolach: Is it IxException? Maybe change exception handling in java?

ixmp/core.py
```diff
@@ -520,6 +535,15 @@ def add_timeseries(self, df, meta=False):
...
         df.columns = df.columns.astype(int)
 
+        # Identify columns to drop
+        to_drop = set()
+        if year_lim[0]:
+            to_drop |= set(filter(lambda y: y < year_lim[0], df.columns))
+        if year_lim[1]:
+            to_drop |= set(filter(lambda y: y > year_lim[1], df.columns))
+
+        df.drop(to_drop, axis=1, inplace=True)
```
zikolach: nit: maybe avoid operation if to_drop is empty?

doc/source/file-io.rst
khaeru: No idea. We never had any tests for that. We can open a new issue to extend the format, i.e. auto-splitting across multiple sheets?

doc/source/file-io.rst
khaeru: That's what has always been returned by `Scenario.equ()` and `Scenario.par()`.

doc/source/file-io.rst
khaeru: I think it *should*, but the data model certainly does *not* current enforce uniformity of units for a given parameter. Mixed units are common in MESSAGEix-GLOBIOM instances; I know because we have to do a lot in the reporting to work around that.

Unless that change is made, this splitting of units to an additional sheet could only be done part of the time. I'm not very interested in making the Excel format more complex, but like https://github.com/iiasa/ixmp/pull/289#discussion_r397446994 an issue could be filed for possible future follow-up.

ixmp/backend/base.py
khaeru: There are no scalar (0-dimensional) sets in the data model. An index set is always 1-D; an indexed set is indexed by at least 1 index set, so it is ≥1-D.

ixmp/core.py
khaeru: I think it's a no-op and not very expensive.

ixmp/backend/jdbc.py
khaeru: Longer-term, I think it is best to specify on the Python side, i.e. through "Raises" sections in the base.Backend docstrings, and through tests, what happens for each method if something goes wrong.

When touching existing code, I think, "What about a hypothetical Backend that's *not* backed by a database?" Ideally, the same exceptions—type and message—should be raised by this hypothetical Backend and by JDBCBackend for the same user error. This means the user can write `try:`/`except…` code that works no matter which Backend is used.

PRs are welcome that change things on the Java side and remove this catch-and-reraise code.

ixmp/backend/jdbc.py
khaeru: Same response as https://github.com/iiasa/ixmp/pull/289#discussion_r397454945.

doc/source/file-io.rst
zikolach: I think there were several parameters having >1M elements, but also potentially with adding subannual feature to model data will significantly increase this number.

doc/source/file-io.rst
zikolach: > Mixed units are common in MESSAGEix-GLOBIOM instances; I know because we have to do a lot in the reporting to work around that.
What do you mean exactly as mixed units? Having more than one unit per parameter or something else?

doc/source/file-io.rst
khaeru: Created #292 in case anyone has such proposals.

doc/source/file-io.rst
khaeru: Yes, "mixed units" means having more than one unit per parameter.

ixmp/backend/jdbc.py
```diff
@@ -603,13 +642,10 @@ def init_item(self, s, type, name, idx_sets, idx_names):
...
+                raise ValueError(f'{name!r} already exists')
```
zikolach: https://www.python.org/dev/peps/pep-0498/#s-r-and-a-are-redundant

ixmp/backend/jdbc.py
zikolach: What I meant is - to avoid relying on specific exception messages. If needed to distinguish the cause we may (a) introduce error codes or (b) create more exception types in java (e.g. IxNotFoundException or similar).
Relying on exception message is often fragile.

ixmp/backend/jdbc.py
khaeru: Ah, nice. We should then have an issue like pandas-dev/pandas#29886 and do all ~50 usages at once.

ixmp/backend/jdbc.py
khaeru: Separate PRs would also be welcome to do that!

ixmp/backend/jdbc.py
khaeru: Created #293.

ixmp/backend/base.py
zikolach: Sorry, I didn't see at first it is changing description of one of return types

ixmp/backend/jdbc.py
zikolach: Is it ok if we create separate issue? I had a quick look on existing code and it seems to be common pattern in many places to parse error messages.

khaeru: Created #296.

PR #290, opened 2020-03-23
Try to fix failing build on AppVeyor

zikolach:
- remove finally block in timeseries commit method
- change input validation in setStatus method
- fix commit in Scenario class
- propagate exceptions from commit
- upgrade pytest dependency
- tee stdout on appveyor


khaeru:
[AppVeyor build 517 passed](https://ci.appveyor.com/project/iiasa/ixmp/builds/31662587). The logs are, I guess as expected, very long.

@zikolach can you look through and collect any noteworthy information, then try removing the `--capture=tee-sys` flag to see if they still pass? https://github.com/iiasa/ixmp/blob/78628a6610897a78e6d2d1ca6f3590bf978f4f07/.appveyor.yml#L40

After that we can merge and continue with #289.

.appveyor.yml
```diff
@@ -37,4 +37,4 @@ build: false
 
 test_script:
 # Run tests. These include the equivalent of Travis R tests via test_r.py.
-- pytest --test-r --verbose
+- pytest --test-r --verbose --capture=tee-sys
```
khaeru: Is it preferable to leave this on permanently? Could you explain why?

ixmp/backend/base.py
```diff
@@ -181,6 +181,18 @@ def set_log_level(self, level):
...
+
+        See also
+        --------
+        get_log_level
+        """
+
+    def get_log_level(self):
+        """OPTIONAL: Get logging level for the backend and other code.
+
+        See also
```
khaeru:
```
Returns
-------
str
    Name of a :ref:`Python logging level <py:levels>`.
```
if that's the case.

ixmp/core.py
```diff
@@ -111,6 +111,12 @@ def set_log_level(self, level):
         logger().setLevel(level)
         self._backend.set_log_level(level)
 
+    def get_log_level(self):
+        """Get global logger level.
```
khaeru: "Get a log level associated with the storage :class:`.Backend`, if any."

Also add the "Returns" section as above.

.appveyor.yml
zikolach: removed it

PR #295, opened 2020-03-27
Add implementation of `timeseries(subannual)` and tests

danielhuppmann:
This PR adds an option for `Scenario.timeseries()` to return the subannual-column

This will solve the issue reported in https://github.com/iiasa/message_ix/issues/322.

## How to review

- Review the extended docstring of the `timeseries()` function and implementation
- View changes in the tests and note that the CI checks all pass.
- Ensure that changes/additions are self-documenting, i.e. that another
  developer like the reviewer will be able to understand what the code does
  in the future.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

closes https://github.com/iiasa/message_ix/issues/322


danielhuppmann:
@zikolach @khaeru finally good to be reviewed

ixmp/core.py
```diff
@@ -583,9 +592,20 @@ def timeseries(self, region=None, variable=None, unit=None, year=None,
...
+                msg = ("timeseries data has subannual values, ",
+                       "use `subannual=True or 'auto'`")
+                raise ValueError(msg)
+            if not has_subannual:
+                df.drop('subannual', axis=1, inplace=True)
+
         if iamc:
             # Convert to wide format
-            index = IAMC_IDX + ['subannual']
+            index = IAMC_IDX + ['subannual'] if 'subannual' in df.columns \
```
zikolach: nit: a bit hard to read as it needs to think/recall [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence). I would do it more explicit.

ixmp/core.py
```diff
@@ -583,9 +592,20 @@ def timeseries(self, region=None, variable=None, unit=None, year=None,
         df['model'] = self.model
         df['scenario'] = self.scenario
 
+        # drop `subannual` column if not requested (False) or required ('auto')
+        if subannual is not True:
+            has_subannual = any(df['subannual'].unique() != ['Year'])
```
zikolach: nit: not sure but maybe
```
df[df['subannual'] != 'Year'].any()
```
is cheaper? @khaeru do you have an idea?

ixmp/core.py
danielhuppmann: changer for a more intuitive implementation

ixmp/core.py
danielhuppmann: I doubt that the alternative would be much more efficient, so I'd keep the current implementation because it is easier to read:

> any of df-subannual not equals year

compared to

> df of df-subannual not equals year - and now any of that...

khaeru:
```python
In [1]: s = pd.Series(list(range(1000)), dtype='str')
In [2]: s[500] = 'Year'
In [3]: df = pd.DataFrame(dict(value=123., subannual=s, unit='kg'))
In [4]: %timeit any(df['subannual'].unique() != ['Year'])
139 µs ± 2.67 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [5]: %timeit df[df['subannual'] != 'Year'].any()
807 µs ± 7.21 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [6]: %timeit (df['subannual'] != 'Year').any()
163 µs ± 6.84 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [7]: %timeit set(df['subannual'].unique()) > {'Year'}
157 µs ± 3.73 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [8]: %timeit any(df['subannual'].unique() != 'Year')
142 µs ± 6.76 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [9]: %timeit (df['subannual'] != 'Year').any()
184 µs ± 10.7 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [10]: %timeit not (df['subannual'] == 'Year').all()
155 µs ± 3.58 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [11]: %timeit ts = df['subannual'].unique(); len(ts) == 1 and ts[0] == 'Year'
111 µs ± 818 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [12]: %timeit not all(df['subannual'] == 'Year')  
116 µs ± 3.39 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

There's not a lot of difference, indeed. For readability, I would slightly prefer the last one, which is almost plain English: "not all subannual values are 'Year'"

danielhuppmann: changed to most-close-to-spoken as suggested - thanks for coming up with a dozen alternatives, very impressive...

any idea why only one approach is so much less efficient?

khaeru: > any idea why only one approach is so much less efficient?

5 starts by computing a pd.Series of True/False: `df['subannual'] != 'Year'`. Then it uses that Series to index the DataFrame: `df[...]`. This operation creates a whole new pd.DataFrame (or view) with only the rows that have True in the first step. It's manipulating _all_ the columns, which I guess is why it's slower.

zikolach: @khaeru thanks a lot for comparison!
You are right `df[df['subannual'] != 'Year'].any()` has a mistake, when suggested I meant `(df['subannual'] != 'Year').any()`.
I thought it could be difference between python function `all`/`any` and methods `all`/`any` of Pandas, but it doesn't seem to be the case.
Here are some results for different dataframe sizes:
index

danielhuppmann:
Not sure why codecov complains here - all new lines of code are tested. So I guess this is good to be merged anyway @khaeru...

khaeru:
> So I guess this is good to be merged anyway @khaeru...

Sorry, I kept coming back and seeing the "Release notes updated" box in the description wasn't checked; so I assumed it was still in progress and didn't look further. But I see now that they actually have been updated; so I've checked the box and will review.

danielhuppmann:
merging as approved by @khaeru & @zikolach

PR #301, opened 2020-04-01
Fix access test by increasing mock server wait time

zikolach:


khaeru:
Added the 'ci' label and assignee.

Issue #292, opened 2020-03-25
Handle items with >10⁶ elements in Excel data I/O


khaeru:
In https://github.com/iiasa/ixmp/pull/289#discussion_r397426609, @zikolach pointed out that some Scenarios may have parameters with more than 1 million elements, that might exceed the size limit of the Microsoft Excel format or the Python tools used to read/write it.

`Scenario.{read,write}_excel()` and the associated format should be:
- Tested for their behaviour with such 'large' data, and
- Extended to support round-trip write and read of such data.

khaeru:
Further suggestion from @zikolach at https://github.com/iiasa/ixmp/pull/289#pullrequestreview-381275785:

> In general would be good to allow splitting items data into multiple sheets as we can reach rows limit. We could avoid to have separate sheet with item name-type mapping by embedding type into sheet name, e.g. type|name|n or similar convention, where type is item type, name - item name and n is a number of sheet for "large" items.

That sounds like a good proposal.

Users may have old Excel spreadsheets sitting around and will expect them to continue working. So we would need to add a second code path for these alternate formats; support both for some time; and then deprecate & eventually remove the current one. We would need an explicit decision about when the current format will no longer be supported.

PR #309, opened 2020-04-10
Extend Excel write/read to handle large data beyond the row limits

behnam-zakeri:
In response to issue #292, this PR extends the code in `ixmp.backend.io.py` to write and read large data, i.e., dataframes and series longer than the maximum row number of Excel.

## How to review
You can test this branch as it is. The tests are improved to reflect the new changes. ALternatively, you can export one of the global scenarios and check if parameter _land_output_ is completely transferred to Excel in two sheets.

## PR checklist
- [x] Tests improved to reflect the enhancements.
- [x] Documentation is not needed, minor enhancement.
- [x] Release notes updated by adding a short note.

### DIFF
--- a/ixmp/backend/io.py
+++ b/ixmp/backend/io.py
@@ -2,7 +2,6 @@ from collections import deque
 import logging
 
 import pandas as pd
-import numpy as np
 
 from ixmp.utils import as_str_list
 from . import ItemType
@@ -92,24 +91,20 @@ def s_write_excel(be, s, path, item_type, max_row=None):
                 empty_sets.append((name, data))
             continue
 
-        if len(data) > max_row:
-            for sheet_num in range(1, int(np.ceil(len(data) / max_row)) + 1):
-                first_row = (sheet_num - 1) * max_row
-                last_row = min(max_row * sheet_num, len(data))
-                if isinstance(data, pd.Series):
-                    part = data.loc[first_row:last_row]
-                else:
-                    part = data.iloc[first_row:last_row, :]
+        # Write data in multiple sheets
 
-                if sheet_num > 1:
-                    suffix = '({})'.format(sheet_num)
-                else:
-                    suffix = ''
+        # List of break points, plus the final row
+        splits = list(range(0, len(data), max_row)) + [len(data)]
+        # Pairs of row numbers, e.g. (0, 100), (100, 200), ...
+        first_last = zip(splits, splits[1:])
 
-                part.to_excel(writer, sheet_name=name + suffix,
-                              index=False)
-        else:
-            data.to_excel(writer, sheet_name=name, index=False)
+        for sheet_num, (first_row, last_row) in enumerate(first_last, start=1):
+            # Format the sheet name, possibly with a suffix
+            sheet_name = name + (f'({sheet_num})' if sheet_num > 1 else '')
+
+            # Subset the data (only on rows, if a DataFrame) and write
+            data.iloc[first_row:last_row] \
+                .to_excel(writer, sheet_name, index=False)
 
     # Discard entries that were not written
     for name in omitted:
--- a/ixmp/backend/io.py
+++ b/ixmp/backend/io.py
@@ -2,6 +2,7 @@ from collections import deque
 import logging
 
 import pandas as pd
+import numpy as np
 
 from ixmp.utils import as_str_list
 from . import ItemType
@@ -35,7 +36,7 @@ def ts_read_file(ts, path, firstyear=None, lastyear=None):
     ts.commit(msg)
 
 
-def s_write_excel(be, s, path, item_type):
+def s_write_excel(be, s, path, item_type, max_row=1e6):
     """Write *s* to a Microsoft Excel file at *path*.
 
     See also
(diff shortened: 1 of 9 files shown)


ixmp/backend/io.py
```diff
@@ -84,7 +88,23 @@ def s_write_excel(be, s, path, item_type):
                 empty_sets.append((name, data))
             continue
 
-        data.to_excel(writer, sheet_name=name, index=False)
+        if len(data) > max_row:
+            for i in range(1, int(np.ceil(len(data) / max_row)) + 1):
```
zikolach, line 92: Can you name I as sheet_num or something meaningful?

ixmp/backend/io.py
```diff
@@ -84,7 +88,23 @@ def s_write_excel(be, s, path, item_type):
...
+                last_row = min(max_row * i, len(data))
+                if isinstance(data, pd.Series):
+                    part = data.loc[(i - 1) * max_row:last_row]
```
zikolach, line 95: Maybe calculate first row index before if block

ixmp/backend/io.py
```diff
@@ -61,6 +62,9 @@ def s_write_excel(be, s, path, item_type):
     omitted = set()
     empty_sets = []
 
+    # Don't allow the user to set a value that will truncate data
+    max_row = int(min([x for x in [max_row, EXCEL_MAX_ROWS] if x is not None]))
```
zikolach, line 66: Is it just to calc min of 2?

ixmp/backend/io.py
```diff
@@ -84,7 +88,23 @@ def s_write_excel(be, s, path, item_type):
...
+                last_row = min(max_row * i, len(data))
+                if isinstance(data, pd.Series):
+                    part = data.loc[(i - 1) * max_row:last_row]
+                else:
+                    part = data.iloc[(i - 1) * max_row:last_row, :]
+
+                if i > 1:
+                    suffix = '({})'.format(i)
+                else:
+                    suffix = ''
```
zikolach, line 102: Wouldn't it better to always add numeric prefix if number of elements is more than 1 sheet?

ixmp/backend/io.py
```diff
@@ -176,6 +196,11 @@ def s_read_excel(be, s, path, add_units=False, init_items=False,
         if first_pass:
             # Read data
             data = xf.parse(name)
+            # appending data from repeated sheets due to max row limit
+            sheets = [x for x in xf.sheet_names if (name + '(') in x]
```
zikolach, line 200: Replace in with startswith

ixmp/backend/io.py
```diff
@@ -235,6 +260,11 @@ def s_read_excel(be, s, path, add_units=False, init_items=False,
         # Only parameters beyond this point
 
         df = xf.parse(name)
+        # appending data from repeated sheets due to max row limit
+        sheets = [x for x in xf.sheet_names if (name + '(') in x]
```
zikolach, line 264: Same as before + next if might be redundant

ixmp/cli.py
```diff
@@ -140,9 +140,11 @@ def config(action, key, value):
 
 
 @main.command()
+@click.option('--max-row', type=int,
+              help='Max row numbers in each sheet.')
```
zikolach, line 144: What is the purpose of parameter?

khaeru:
In addition to the points raised by @zikolach, the documentation (current version shown here: https://message.iiasa.ac.at/projects/ixmp/en/master/file-io.html#scenario-model-data) needs to be adjusted to describe the new format.

ixmp/backend/io.py
behnam-zakeri: Thanks, renamed to _sheet_num_

ixmp/backend/io.py
behnam-zakeri: done.

ixmp/backend/io.py
behnam-zakeri: @zikolach, this is needed because the input argument may be None passed through other files (core.py and base.py).

zikolach: I see, just wondered if it can be expressed simpler ;)

ixmp/backend/io.py
behnam-zakeri: Now, the sheets will be saved as foo, foo(1), foo(2), ... . I think it's better to see the sheets starting with the name of the items (in case the sheet name is not fully expanded).

ixmp/backend/io.py
behnam-zakeri: Thanks, improved.

ixmp/backend/io.py
behnam-zakeri: Thanks, the startswith() used. But I believe the next if clause is needed because sheets maybe an empty list, then the following for loop raises an error. Was this your point?

ixmp/cli.py
behnam-zakeri: This is to let the input argument of max_row being passed to to_excel().

behnam-zakeri:
> In addition to the points ... needs to be adjusted to describe the new format.

Thanks @khaeru, I extended the doc to reflect this. I also noticed a redundant sentence and deleted that.

ixmp/backend/io.py
zikolach: which error?

ixmp/cli.py
zikolach: I meant what is the need of max_row parameter in principle? do you want to limit a number of elements in parameter, e.g. save only parameters up to 500 elements - something like this?

ixmp/backend/io.py
behnam-zakeri: You are right, I removed the if clause. thanks.

ixmp/backend/io.py
khaeru: I modified as follows:
https://github.com/iiasa/ixmp/blob/96528d8c405e49e6da00d2f6afe5603b9e2b5fbb/ixmp/backend/io.py#L68-L69
1. If `max_rows` is None, then `max_rows or EXCEL_MAX_ROWS` evaluates to the latter.
2. Call `int()` on (1)
3. Then the `min()` of (2) and `EXCEL_MAX_ROWS`, i.e. possibly the min of two identical values.

ixmp/cli.py
khaeru: I don't think there's any harm in adding this!

behnam-zakeri:
Thanks @zikolach for useful comments and @khaeru for very nice improvements of the code.

PR #313, opened 2020-04-14
Always fetch the pint 'application registry' from within functions

khaeru:
- As a follow-up to #312, switch to an implementation where `pint.get_application_registry()` is always called from within functions. This allows downstream code to replace this registry.
- Add a comment explaining this choice for future developers.

## PR checklist

- [x] ~Tests added.~ N/A
- [x] ~Documentation added.~
- [x] ~Release notes updated.~


khaeru:
The codecov/patch check failure is spurious.

PR #314, opened 2020-04-16
Bulk saving for metadata and exposing documentation API

zikolach:
Allow to update multiple metadata categories at once.

**Important**: based on PR iiasa/ixmp_source#290

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


ixmp/backend/jdbc.py
```diff
@@ -802,15 +857,31 @@ def unwrap(v):
         return {entry.getKey(): unwrap(entry.getValue())
                 for entry in self.jindex[s].getMeta().entrySet()}
 
-    def set_meta(self, s, name, value):
+    def set_meta(self, s, name_or_data, value=None):
```
khaeru, line 860: Is this still consistent with `base.Backend`?

ixmp/backend/jdbc.py
```diff
@@ -802,15 +857,31 @@ def unwrap(v):
...
+        if type(name_or_data) == dict:
+            name_or_data = list(name_or_data.items())
```
khaeru, line 862: These two lines only seem like something that would be done to any user input, regardless of which Backend is in use. Is that accurate? If that's so, then it should be done in core.py.

ixmp/backend/jdbc.py
```diff
@@ -802,15 +857,31 @@ def unwrap(v):
...
+        if type(name_or_data) == list:
+            jdata = java.LinkedHashMap()
+            for k, v in name_or_data:
+                jdata.put(str(k), v)
+            self.jindex[s].setMeta(jdata)
+            return
         _type = type(value)
```
khaeru, line 869: Add a blank line before continuing after an early `return`.

ixmp/backend/jdbc.py
```diff
@@ -802,15 +857,31 @@ def unwrap(v):
...
         try:
             _type = {int: 'Num', float: 'Num', str: 'Str', bool: 'Bool'}[_type]
             method_name = 'setMeta' + _type
         except KeyError:
             raise TypeError(f'Cannot store metadata of type {_type}')
 
-        getattr(self.jindex[s], method_name)(name, value)
+        getattr(self.jindex[s], method_name)(name_or_data, value)
+
+    def delete_meta(self, s, name_or_names):
```
khaeru, line 878: Is this in `base.Backend`?

ixmp/core.py
```diff
@@ -1502,17 +1504,31 @@ def get_meta(self, name=None):
...
+        self._backend('set_meta', name_or_data, value)
+
+    def delete_meta(self, name_or_names):
+        """Set scenario metadata.
+
+        Parameters
+        ----------
+        name_or_names : str or list
+            If the argument is list, it used to remove multiple metadata
+            entries at once. Otherwise, use the argument
```
khaeru, line 1528: Check wrapping.

zikolach: Not sure what you mean as wrapping

zikolach:
@khaeru thanks for comments! Please have a look at https://github.com/iiasa/ixmp/pull/314/commits/6c37514f48463e9ac6b2841dc59cc04dcf7f121a - it should be fixed there

ixmp/backend/base.py
```diff
@@ -882,15 +882,17 @@ def get_meta(self, s: Scenario):
         """
 
     @abstractmethod
-    def set_meta(self, s: Scenario, name, value):
-        """Set a single metadata key.
+    def set_meta(self, s: Scenario, name_or_data, value=None):
+        """Set single or multiple metadata entries.
```
danielhuppmann, line 886: please rename to `meta` throughout

ixmp/backend/base.py
```diff
@@ -882,15 +882,17 @@ def get_meta(self, s: Scenario):
...
 
         Parameters
         ----------
-        name : str
-            Metadata key name.
-        value : int or float or bool or str
-            Value for *name*.
+        name_or_data : str or dict
```
danielhuppmann, line 890: suggest to call `name_or_dict`, because data is misleading (we are talking about meta indicators, not data in the traditional sense)

ixmp/backend/base.py
```diff
@@ -902,6 +904,20 @@ def set_meta(self, s: Scenario, name, value):
             If *value* is not a valid type.
         """
 
+    @abstractmethod
+    def delete_meta(self, s, name_or_names):
```
danielhuppmann, line 908: no need to call this arg `name_or_names`, `name` is sufficiently clear

ixmp/core.py
```diff
@@ -55,6 +55,8 @@ class Platform:
     _backend_direct = [
         'open_db',
         'close_db',
+        'get_doc',
```
danielhuppmann, line 58: I'm confused, is there now a top-level function to get or set docstrings?

ixmp/backend/jdbc.py
```diff
@@ -802,15 +857,30 @@ def unwrap(v):
         return {entry.getKey(): unwrap(entry.getValue())
                 for entry in self.jindex[s].getMeta().entrySet()}
 
-    def set_meta(self, s, name, value):
+    def set_meta(self, s, name_or_data, value=None):
```
danielhuppmann, line 860: see my comment on the backend

ixmp/backend/jdbc.py
```diff
@@ -226,6 +238,49 @@ def get_log_level(self):
         levels = {v: k for k, v in LOG_LEVELS.items()}
         return levels.get(self.jobj.getLogLevel(), 'UNKNOWN')
 
+    def set_doc(self, domain, docs):
+        """ Save documentation to database
```
danielhuppmann, line 242: is this consistent with space after `"""` or without?

ixmp/backend/jdbc.py
```diff
@@ -226,6 +238,49 @@ def get_log_level(self):
...
+        self.jobj.setDoc(dd, jdata)
+
+    def get_doc(self, domain, name=None):
+        """ Read documentation from database
+
+        domain : str
+            Documentation domain, e.g. model, scenario etc
+        name : str, optional
+            Name of domain entity (e.g. model name).
+        Returns
```
danielhuppmann, line 268: extra line missing?

ixmp/backend/jdbc.py
```diff
@@ -226,6 +238,49 @@ def get_log_level(self):
...
+        if type(docs) == dict:
+            docs = list(docs.items())
+        for k, v in docs:
+            jdata.put(str(k), str(v))
+        self.jobj.setDoc(dd, jdata)
+
+    def get_doc(self, domain, name=None):
+        """ Read documentation from database
+
+        domain : str
```
danielhuppmann, line 264: `Parameters` missing?

ixmp/core.py
danielhuppmann: Ok, thanks for the clarification - how is this rendered in the docs? Is there an example where I can see this already?

zikolach: Good question! I think it is not available on [Platform class documentation](https://message.iiasa.ac.at/projects/ixmp/en/stable/api-python.html#platform).

@khaeru what do you think, can we adjust `__doc__` attribute to incorporate those backend methods documentation into Platform docs?

ixmp/backend/jdbc.py
zikolach: updated

ixmp/backend/base.py
zikolach: updated

ixmp/backend/base.py
zikolach: updated

ixmp/backend/base.py
zikolach: updated (I hope I didn't miss any other places)

ixmp/backend/jdbc.py
zikolach: updated

ixmp/backend/jdbc.py
zikolach: updated

ixmp/backend/jdbc.py
zikolach: updated

zikolach:
>     * Documentation of the Backend API goes on the abstract methods in `base.Backend`. The implementations of these methods on `jdbc.JDBCBackend` generally have no docstring, unless they have significant differences in behaviour from the general API which need explanation. See the existing code.
> 
>     * Please update `doc/source/api-backend.rst` to refer to the new methods.

Thanks for suggestions. I moved docstrings to backend base class and added new methods to `api-backend.rst`.

danielhuppmann:
> I moved docstrings to backend base class and added new methods to api-backend.rst.

Does this mean these functions will not be included in the Platform docs? If that's the case, I don't think that this is a good strategy...

zikolach:
> Does this mean these functions will not be included in the Platform docs

Depend what you mean as Platform docs. These methods docstrings are added to [Backend documentation](https://message.iiasa.ac.at/projects/ixmp/en/stable/api-backend.html#backend-api), not to [Platform class documentation](https://message.iiasa.ac.at/projects/ixmp/en/stable/api-python.html#platform).

danielhuppmann:
This reference to the new functions is not understandable to anyone who was not part of this discussion. 

> The methods `~.base.Backend.open_db`, `~.base.Backend.close_db`, `~.base.Backend.get_doc`, and `~.base.Backend.set_doc` may also be called via Platform.

It's not obvious how to call these functions - `Platform.base.Backend.open_db`?

khaeru:
> This reference to the new functions is not understandable […]
>
>> The methods `~.base.Backend.open_db`, `~.base.Backend.close_db`, `~.base.Backend.get_doc`, and `~.base.Backend.set_doc` may also be called via Platform.

See the [Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax):
> If you prefix the content with `~`, the link text will only be the last component of the target. For example, `:py:meth:'~Queue.Queue.get'` will refer to `Queue.Queue.get` but only display `get` as the link text. 

Sphinx renders the sentence as "The methods `open_db()`, `close_db()`, …" with links.

If that doesn't suit a particular use case/target user/story, other alternatives are:
- Use the [Sphinx `.. py:function::` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-function) to put the user-facing documentation in api-python.rst.
- Don't use _backend_direct (originally introduced as a convenience when developing the Backend API), and write short (possibly one-line) methods. See for example `add_unit`/`units`, `set_log_level`/`get_log_level`. In this case, observe separation of concerns [as described in the docs](https://message.iiasa.ac.at/projects/ixmp/en/latest/api-backend.html#backend-api); e.g. just as `set_log_level` raises ValueError for invalid input, `set_doc` should raise ValueError for an invalid domain.

PR #190, opened 2019-09-24
Disable Travis caching of Miniconda to resolve CI errors

khaeru:
Travis builds (e.g. [#547](https://travis-ci.org/iiasa/ixmp/builds/574821631)) are erroring with messages like:
```
Downloading and Extracting Packages

python-libarchive-c- | 23 KB     |            |   0% 
conda-package-handli | 255 KB    |            |   0% 
bzip2-1.0.8          | 75 KB     |            |   0% 
sqlite-3.29.0        | 1.2 MB    |            |   0% 

RuntimeError('EnforceUnusedAdapter called with url https://repo.anaconda.com/pkgs/main/osx-64/python-libarchive-c-2.8-py37_11.conda\nThis command is using a remote connection in offline mode.\n',)
RuntimeError('EnforceUnusedAdapter called with url https://repo.anaconda.com/pkgs/main/osx-64/conda-package-handling-1.3.11-py37_0.conda\nThis command is using a remote connection in offline mode.\n',)
RuntimeError('EnforceUnusedAdapter called with url https://repo.anaconda.com/pkgs/main/osx-64/bzip2-1.0.8-h1de35cc_0.conda\nThis command is using a remote connection in offline mode.\n',)
RuntimeError('EnforceUnusedAdapter called with url https://repo.anaconda.com/pkgs/main/osx-64/sqlite-3.29.0-ha441bb4_0.conda\nThis command is using a remote connection in offline mode.\n',)

ci/travis-install.sh: line 10: conda: command not found
```
… etc.

This PR is to debug/resolve.

**PR checklist:**

- [x] ~Tests added.~ Not applicable; this is not a user-facing/API change.
- [x] ~Documentation added.~
- [x] ~Release notes updated.~

PR #315, opened 2020-04-23
Add precautionary checks to Model.initialize_items

khaeru:
This PR adds changes necessary for iiasa/message_ix#190. In particular, it ensures that the `Model.initialize()` method is *always* called when a Scenario is cloned *or* loaded.

This helps in the following way:
- In that PR, `MESSAGE.initialize()` is expanded to add sets and parameters related to storage to any existing Scenario that lacks them. E.g. in `test_legacy_version`, an old Scenario is loaded that lacks the storage parameters. Solving that Scenario with the updated MESSAGE GAMS code fails, because the items expected are not found.
- The code in this PR ensures that `MESSAGE.initialize()` is invoked when that Scenario is loaded: https://github.com/behnam-zakeri/message_ix/blob/43a98fe3e416bf1fe61af54f3243052dc162f23d/message_ix/tests/test_legacy_version.py#L15-L16

Also:
- `Model.initialize_items()` checks whether the returned object is in a checked-out state when it is received, and leaves the object in the same state.
- `Backend.init_ts()` and `Backend.init_s()` are consolidated and simplified.
- `Scenario.__init__()` now chains through its parent method, `TimeSeries.__init__()`.
- Made some follow-up changes to #320. `test_show_versions` would load pyam and message_data, which both change the pint 'application registry'; this would cause the ixmp reporting tests related to units to fail, depending on which subset of tests was run.

## How to review

- Note that CI passes.
- Run with iiasa/message_ix#190 and confirm the message_ix test `test_legacy_version` passes.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -4,6 +4,7 @@ Next release
 All changes
 -----------
 
+- `#315 <https://github.com/iiasa/ixmp/pull/315>`_: Ensure :meth:`.Model.initialize` is always called for new *and* cloned objects.
 - `#320 <https://github.com/iiasa/ixmp/pull/320>`_: Add CLI command `ixmp show-versions` to print ixmp and dependency versions for debugging.
 - `#312 <https://github.com/iiasa/ixmp/pull/312>`_: Add :meth:`~.computations.apply_units`, :meth:`~computations.select` reporting calculations; expand :meth:`.Reporter.add`.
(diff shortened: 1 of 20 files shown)


behnam-zakeri:
Thanks @khaeru for imprving the code to account for temporary and/or new items introduced in a development branch. I see this very useful, as the user can add these new/temporary items in one place `message_ix.models.py` and make sure that the entire ixmp-message_ix ecosystem will handle these items as expected. I tested this in combination with [#190](https://github.com/iiasa/message_ix/pull/190) in message_ix and it works well.

Issue #274, opened 2020-03-05
Harmonize handling of codelists


khaeru:
At the 2020-03-04 MESSAGE meeting, we discussed that several items in the [ixmp data model](https://github.com/iiasa/ene/wiki/Data-model) are actually instances of a common class. Namely:

- ‘unit’
- ‘region’ (name used for for the code list entries) / ‘node’ (name used for data dimensions).
- and now, per #264, ‘timeslice’ (name used for the code list entries) / ‘subannual’ (name used for data dimensions).

I propose to:

- Simplify the Backend interface (see [here](https://message.iiasa.ac.at/projects/ixmp/en/stable/api-backend.html#ixmp.backend.base.Backend)) as follows. These changes are invisible to the user; they just affect how a Platform talks to a Backend.
  - Add methods:
    - `get_codes(codelist)`
    - `set_code(codelist, id, name=None, description=None, **attrs)`

    Valid values will be:
    - for the *codelist* argument: 'node', 'timeslice' (or 'time_slice'?), 'unit'.
    - for *attrs*:
      - codelist='node': 'parent', 'hierarchy', 'mapped_to'.
      - codelist='timeslice': 'category', 'duration'.
      - codelist='unit': none.
  - Delete methods: `get_nodes`, `get_timeslices`, `get_units`, `set_node`, `set_timeslice`, `set_unit`.
- Add to the Platform (without removing any existing behaviour):
  - an attribute like `codelist`, so that:
    - `mp.codelist['region']` is the same as `mp.get_regions()`, etc.
  - A method `add_code(...)` with the same signature as `Backend.set_code()`.

This should be addressed after #264 is completed and merged.

*Rationale:* The proliferation of different types of objects in the data model makes it more costly to develop other, entirely open source Backends as alternatives to JDBCBackend, as targeted in #78. This change reduces the number of distinct methods by handling lists of codes in a consistent way.

khaeru:
- Updated the description as discussed at https://github.com/iiasa/ixmp/pull/353#issuecomment-672764759.
- Note that #322, handling 'docs' and 'meta(data)' as annotations, was not merged with #314, but the branch and code are retained as examples for a future PR. That change is separate from the one described in this issue.

PR #322, opened 2020-04-28
Handle 'annotations' as abstraction of 'docs' and 'metadata'

khaeru:
This PR:

- Modifies the base.Backend API:
  - Add Backend.{set,get,delete}_anno and .get_codes.
  - Remove Backend.{set,get,delete}_meta.
- Implements the new methods in JDBCBackend.
  - Deletes new methods JDBCBackend.{set,get}_doc that were in this class but not added to the base.Backend API.
- Updates Scenario.{get,set,delete}_meta.
- Adds Platform methods:
  - annotate, get_annotations, delete_annotations —for a single targets and 1+ annotations.
  - {set,get}_doc —convenience methods for setting the 'doc' annotation of multiple targets.

This approach:
- Achieves a superset of the functionality targeted by #314 and the same UX.
- Keeps the Backend API and data model smaller and simpler:
  - +0 methods instead of +2.
  - Uses common semantics for annotations (even though these are presented to the user with a variety of names), instead of introducing slightly different semantics for 'metadata' and 'documentation'.
- Takes us closer to, rather than further from:,
  -  #274—by adding `.get_codes()` and `backend.CodeList` to make explicit the commonality between different lists of 'codes' (string IDs).
  - SDMX compatibility—by making clear that 'metadata' and 'documentation' are particular applications of annotations.

Some things that could be improved, now or after this is merged into the branch for #314:
- JDBCBackend.get_codes can't get a list of existing 'metadata categories' that might be annotated.
- JDBCBackend.get_codes has to use several lines to get a list of run IDs, since I couldn't find a public method in the Java code to return this.
- CodeList.variable corresponds to the 'timeseries' member of the Java DocumentationDomain enumeration, so jdbc._domain_enum() has to remap this. I think the latter is a misnomer, and uses the phrase 'time series' in a place when it's already used twice (TimeSeries class, and 'time series data' stored in a TimeSeries object). These 'doc' annotations are actually attached to *(fragments of) values occurring in the 'variable' dimension of 'time series data'*, hence the chosen name.


ixmp/backend/__init__.py
```diff
@@ -20,6 +20,27 @@
 BACKENDS = {}
 
 
+class CodeList(Enum):
+    """Lists of codes in the ixmp data model."""
+    #: Annotation IDs. See :meth:`.set_anno`
+    metadata = auto()
```
danielhuppmann, line 26: please change to `meta` as described in the data model

ixmp/backend/jdbc.py
```diff
@@ -849,38 +810,80 @@ def item_delete_elements(self, s, type, name, keys):
...
+            target = self.jindex[target_id]
 
-    def set_meta(self, s, name_or_data, value=None):
-        if type(name_or_data) == list:
-            jdata = java.LinkedHashMap()
-            for k, v in name_or_data:
-                jdata.put(str(k), v)
-            self.jindex[s].setMeta(jdata)
-            return
+            # 'metadata'
```
danielhuppmann, line 819: please change to `meta`

PR #332, opened 2020-06-08
Add .eggs files to .gitignore

francescolovat:
This PR addresses the same issue as in https://github.com/iiasa/message_ix/issues/355.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 *#
 *.Rproj
 *.Rcheck
+*.eggs
 
 # Apple file system
 .DS_Store


francescolovat:
Hi @zikolach,

do you know why `test_check_single_model_access` is failing? It is passing in my IIASA PC 😞
Should I just try to re-run the build?

https://ene-builds.iiasa.ac.at/viewLog.html?buildId=4124&buildTypeId=IxmpCore_IxmpPython&tab=buildLog&_focus=437

khaeru:
>  Should I just try to re-run the build?

I already restarted this, the build is queued. This test is "flaky": it randomly fails from time to time. That's because it uses "pretenders" to start a fake HTTP server, and then it makes some requests against that server. Sometimes the server fails to start at all, or quickly enough.

PR #340, opened 2020-06-19
Use rcmdcheck::rcmdcheck() to run R checks and tests

### LINKED ISSUE #328: Address R testthat incompatibility; re-enable rixmp tests
- r-lib/testthat#1035 was identified via #279. Code which runs from Python, or from R, causes segfaults when run within an R 'testthat' test suite. The cause lies somewhere along the R–reticulate–Python–JPype–ixmp_source chain.
- As a temporary mitigation, the rixmp tests were temporarily disabled in #327.
- As soon as the upstream issue is diagnosed and a fix identified, the R tests should be re-enabled.
- In the meantime, manual testing of rixmp can be done by @adrivinca @guofei2016, and other R users.

khaeru:
This improves the R test suite so that `rcmdcheck::rcmdcheck()` can be invoked as part of the GitHub Actions workflow. R test are re-enabled on CI, closing #328.

The only substantive changes:
- test_r.py ([final version](https://github.com/iiasa/ixmp/blob/98c00a93b0bd10b89c5d1262d9d5cfde6f7d044a/ixmp/tests/test_r.py)) is retired.
  - This allowed pytest to invoke the R testthat tests, but also required arcane adjustments to link the environment between the two; these are now obviated, reducing maintenance workload for the test suite and CI.
  - The IR/Jupyter tutorial notebooks are still tested through the pytest suite.
- In JDBCBackend. To summarize:
  - The return values for `Scenario.var()` or any scalar item were coming back as JPype/Java types, rather than Python `float()`.
  - Under the Python principle of duck typing (“if it walks like a duck, and quacks like a duck…”) these were fine for our entire Python test suite. But reticulate was unable to handle them, which caused failures.
  - Now they are explicitly cast to Python float, so that reticulate can properly handle them.

Similar changes could be applied in message_ix to simplify/expand testing.

## How to review

Note that CI checks pass.

## PR checklist

- [x] Tests added.
- [x] ~Release notes updated.~ N/A, not a user-facing change.


khaeru:
One of the R tests, roughly:
```R
test_that('the canning problem can be solved', {
  # Create the Scenario
  mp <- test_mp()

  scen <- ixmp$testing$make_dantzig(mp, solve = TRUE)

  # Check value
  print(waldo::compare(tmp, 153.675))
})
```
fails (see [here](https://github.com/iiasa/ixmp/pull/340/checks?check_run_id=932044064)) with a message like:
```
[1] "3"
`x` is an S3 object of class <jpype.types.JFloat/_jpype._JNumberFloat/python.builtin.float/_jpype._JObject/python.builtin.object>
`y` is a double vector (153.675)
```

This indicates that JDBCBackend is leaking JPype/Java objects through to the Python API and thus rixmp, when it should only return Python types. Will try to correct this.

PR #345, opened 2020-07-07
Correct error in io.s_read_excel for items on multiple sheets

behnam-zakeri:
This PR resolves minor issues that triggers an error when reading data from Excel for parameters with large data that have been saved in multiple sheets (e.g., foo(1), foo(2)). The concatecation of datafrmaes should be done along axis=0.
The reviewer needs to run this over existing tests. Alternatively, a large MESSAGE parameter saved over two or more Excel sheets should be imported without an error.

- [x] Tests already exists. They need to be improved not to pass if concatecation is done along columns.
- [x] Documentation not needed.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -8,17 +8,11 @@ All changes
 - :pull:`363`: Expand documentation and revise installation instructions.
 - :pull:`362`: Raise Python exceptions from :class:`.JDBCBackend`.
 - :pull:`354`: Add :meth:`Scenario.items`, :func:`.utils.diff`, and allow using filters in CLI command ``ixmp export``.
-- :pull:`353`: Add meta functionality.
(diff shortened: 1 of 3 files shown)


khaeru:
@behnam-zakeri good find. To create the test, I would suggest using the `ixmp.testing.add_random_model_data()` function we added for testing performance on "big" scenarios: https://github.com/iiasa/ixmp/blob/ddbe600394fcbf5edce4d7371bd4b334ecd85a9e/ixmp/testing.py#L633-L647

So your test would have the following steps:
1. Create an empty Scenario.
2. Add data with `length=EXCEL_MAX_ROWS + 1`.
3. Write to file.
4. Read.

Step (4) should fail on the concat without your fix, but pass with it.

If necessary, please delegate the task of writing the test to a teammate.

khaeru:
Since this seems stalled, I'll step in to write the required test(s).

khaeru (approved):

behnam-zakeri:
@khaeru thanks for correcting this and adding the test.

Issue #30, opened 2018-05-08
Add feature to unlock `Timeseries`/`Scenarios` in a database


danielhuppmann:
When checking out an `ixmp.Scenario` and not committing or discarding (or running into an error), the run-id is locked in the database instance. We need an elegant way to 'unlock' the run-id.

Follow-up: this should only be available to admins on central/shared database instance?

danielhuppmann:
A current workaround is to open the underlying database and, in table `RUN`, set the `status` column in the row for the `id` to `1`.

khaeru:
#30 was reported as a duplicate. I left [this comment](https://github.com/iiasa/ixmp/issues/234#issuecomment-564538022), copied below to consolidate.

[Backend API docs](https://message.iiasa.ac.at/projects/ixmp/en/master/api-backend.html#ixmp.backend.base.Backend), for reference.

Some thoughts:
- The `.check_out()`/`.commit()` methods are in TimeSeries, rather than Scenario.
- I've always felt these were a weird hybrid of SQL and git semantics.
  - In SQL, there is nothing like `CHECK_OUT`, only `COMMIT`, `ROLLBACK`, and some others.
    - ixmp lacks anything like `ROLLBACK`.
  - `git checkout` modifies or *reverts* the working tree.
    - In most workflows, it is not needed to call it before modifying files.
    - Often, when used, its meaning is actually closer to SQL `ROLLBACK`: `git checkout -- file.txt` rolls back file.txt to its content as of the last commit.
  - Neither includes a notion of locking.
- `.check_out()` and `.commit()` actually do THREE things:
  - Manage transactions.
  - Locking.
  - `.check_out()` raises an exception if `.has_solution()` is True. (cf. #113 under “Methods in the wrong place”)
- The 'locked' status of a TimeSeries seems to mean 'locked for other users'.
  - ixmp.Platform and base.Backend don't explicitly discuss the (possibly) multi-user nature of Backends.
  - In SQL, if a transaction fails or is rolled back, it is simply not committed, and nothing prevents another set of modifications from starting immediately.

zikolach:
> ixmp lacks anything like ROLLBACK

there is discard_changes method

> git checkout modifies or reverts the working tree

It is incorrect comparison, the semantics of checkout in ixmp is locking of scenario version to make changes. More appropriate would be to compare with [svn lock](http://svnbook.red-bean.com/en/1.7/svn.ref.svn.c.lock.html).

[...]

PR #350, opened 2020-07-15
Add read/write TimeSeries.locked attribute

khaeru:
@behnam-zakeri mentioned that he and a YSSP student both had an active need to use the Java method `Timeseries.unlockRunid()`, so here's an attempt to tackle what turns out to be our oldest issue, #30. This PR:
- Adds getter and setter methods for an attribute `TimeSeries.locked`.
- Adds the Backend API methods `{get,set}_locked`.
- Implements these methods on JDBCBackend using existing ixmp_source methods.

In https://github.com/iiasa/ixmp/issues/30#issuecomment-584551801 I commented on the complexity of the multiple state options/transitions for TimeSeries on the Java side, and @zikolach have had some discussions circling around how to simplify them. This PR *does not* try to alter them; however, it does create the appearance (to the user) that there are two distinct, boolean states: `checked_out`, and `locked`.

Two things that require input, probably from @zikolach:
- [x] `set_locked(ts, False)` should be prohibited for the shared production databases on gp3.iiasa.ac.at. Options:
  - Only allow unlocking when `driver='hqsldb'`, i.e. someone is working in a local database, and don't allow unlocking for any `driver='oracle'` database. I'm not sure if this rules out @behnam-zakeri's main use-case—could you comment? This can be done entirely in Python.
  - ~Change code on the Java side. Not knowing that code, I don't know how this could be done most simply, or whether the changes would echo back into the Python side. If it involves adding lots of auth machinery to the Python code, I would prefer to avoid that.~ No input, so did not choose this option.
- [ ] There's currently no way to force, from Python, a TimeSeries to become locked, in order to actually test whether the unlocking feature works. Is it possible to add such a method? We can keep it hidden from the Python API.

## How to review

Note that the CI checks all pass.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -4,6 +4,7 @@ Next release
 All changes
 -----------
 
+- :pull:`350`: Add read/write :attr:`.locked` attribute to :class:`.TimeSeries`/:class:`.Scenario`.
 - :pull:`345`: Fix a bug in :meth:`.read_excel` when parameter data is spread across multiple sheets.
 - :pull:`363`: Expand documentation and revise installation instructions.
 - :pull:`362`: Raise Python exceptions from :class:`.JDBCBackend`.
--- a/doc/source/api-backend.rst
+++ b/doc/source/api-backend.rst
@@ -147,6 +147,7 @@ Backend API
       get
       get_data
       get_geo
+      get_locked
       init
       is_default
       last_update
@@ -155,6 +156,7 @@ Backend API
       set_data
       set_as_default
       set_geo
+      set_locked
 
    Methods related to :class:`ixmp.Scenario`:
 
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -98,6 +98,11 @@ TimeSeries
       set_as_default
       timeseries
 
+   and the following attribute:
+
+   .. autosummary::
+      locked
+
 
 Scenario
 --------
--- a/ixmp/backend/base.py
+++ b/ixmp/backend/base.py
@@ -523,6 +523,39 @@ class Backend(ABC):
         """
 
     @abstractmethod
+    def get_locked(self, ts: TimeSeries):
+        """Return the lock state of *ts*.
+
+        Returns
+        -------
+        bool
+            :obj:`True` if *ts* is locked; :obj:`False` otherwise.
+
+        See also
+        --------
+        set_locked
+        """
+
+    @abstractmethod
+    def set_locked(self, ts: TimeSeries, value: bool):
+        """Set the lock state of *ts*.
+
+        Parameters
+        -------
+        value : bool
+            If :obj:`True`, lock *ts*. If :obj:`False`, unlock *ts*.
+
+        Raises
+        ------
+        NotImplementedError
+            if the requested operation is not supported.
+
+        See also
+        --------
+        get_locked
+        """
+
+    @abstractmethod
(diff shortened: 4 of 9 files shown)


behnam-zakeri:
Thanks @khaeru for the quick reaction and very good background. From the use case view, I have a few points that may be relevant in refactoring this issue in the code:
1. At the moment, the error message the user gets for the following three events is the same, which seems to have different causes, or at least they have different solutions:
a- trying to check out or load a scenario that is already loaded and checked out in one of the open working sessions of the same user.
b- trying to check out a scenario that has got locked in a working session, due to reasons related to communication between the backend and platform. (I don't know the exact cause, but the last time this happened was when a user wanted to use `scenario.add_timeseries()` with region mappings not defined in their platform). 
c- trying to load a scenario that is locked due to being open in a lost/frozen working session. For example, jupyter notebook crashes while the scenario was still checked out.
- The solution for the case _a_ is rather straightforward. The user either do a `scenario.commit() `or `scenario.discard_changes()` in the open working session, and try to load or check out the scenrio in another session.
- For case _b_, the user needs to get the run ID of that scenario and then unlock it:
```
runid = scenario.run_id()
mp._backend.jobj.unlockRunid(runid)
```
- But for the case _c_, the scenario cannot be even loaded in a new session as it is locked. If the user doesn't know the run ID, which is usually the case, it's not clear how to load and unloack this scenario.
2. So far, the examples I mentioned were for one user working on a local DB or on her/his own scenarios in a shared DB. Definitely, for multiple users working on one scenario in a shared DB, it shouldn't be possible to unloack a scenario that is locked by another user in case _a_ and _b_ above, as the first user may continue working on them. But I believe it should be possible, under a protocol, to unloack and open a locked scenario by another user if this is of case _c_.

khaeru:
@behnam-zakeri thanks for those details. Your cases (b) and (c) are difficult to simulate in our test suite; we would have to set up a connection to one platform, force it to crash/freeze _without_ killing the test code, and then try to reconnect to the same database.

(a) might be possible, since it involves making two connections to the same database; I can try that. But I still think, per my second checkbox, that it would be simpler to add a Java-side method that locks the Scenario, and then use that in the Python test suite. So this awaits @zikolach returning from leave.

khaeru:
> (a) might be possible, since it involves making two connections to the same database; I can try that.

It turns out this is not possible. In order to test this feature, we need a Java method to set the locked state of a TimeSeries object.

khaeru:
> we need a Java method to set the locked state of a TimeSeries object.

Requested at iiasa/ixmp_source#333.

zikolach:
> This can be done entirely in Python.


That's incorrect as we use jdbc to interact with DB and locked state is persisted.

zikolach:
> There's currently no way to force, from Python, a TimeSeries to become locked

What does `Timeseries.check_out` do?

khaeru:
This will be revised given the more extensive Java cleanup in iiasa/ixmp_source#334.

khaeru:
> the more extensive Java cleanup in [iiasa/ixmp_source#334](https://github.com/iiasa/ixmp_source/pull/334).

That's turning out to be more extensive than we can handle this week, so postponing this for 3.2.

khaeru:
> This will be revised given the more extensive Java cleanup in iiasa/ixmp_source#334.

Closing until that clean-up is completed; will open a new PR.

PR #351, opened 2020-07-20
Allow passing string to to_jlist (converting to singleton list)

zikolach:
How it was working before?


khaeru:
For posterity: this PR was to address errors like the following, that appeared due to changes between JPype1 0.7.5 and 1.0:
```
TypeError: No matching overloads found for java.util.LinkedList.addAll(str), options are:
        public boolean java.util.LinkedList.addAll(int,java.util.Collection)
        public boolean java.util.LinkedList.addAll(java.util.Collection)
```

PR #352, opened 2020-07-22
typo in docstring reporting/computations/select()

francescolovat:
Small typo in docstring of `ixmp/reporting/computations/select()`.

More typos will be added here in case they will be ecountered.


khaeru:
The check failures are expected/handled in #357, so I will merge this.

PR #353, opened 2020-07-28
Meta support for model/scenario levels

zikolach:
Implement full meta functionality.

Implement full meta functionality on the `Platform` level, and make some adaptions to the `Scenario` meta functionality too. Also, add the possibility of creating a model or scenario on `Platform`. This PR relies on https://github.com/iiasa/ixmp_source/pull/323.

Detailed list of changes:

  - add `Platform.add_model` using `Backend.add_model`
  - add `Platform.add_scenario` using `Backend.add_scenario`
  - add `Platform.models` using `Backend.models`
  - add `Platform.scenarios` using `Backend.scenarios`
  - add `Platform.get_meta` using `Backend.get_meta`
  - add `Platform.set_meta` using `Backend.set_meta`
  - add `Platform.remove_meta` using `Backend.remove_meta`
  - add `Scenario.remove_meta` using `Backend.remove_scenario_meta`
  - deprecate `Scenario.delete_meta`

## How to review

- make sure the tests pass
- check that the built sphinx documentation is how you'd expect it

Problematic things relevant to this PR:
- ambiguous use of the name `Scenario`. As an `ixmp.Scenario` is actually an ixmp_source `run` that has a `scenario` itself, distinguishing between these two types of Scenarios is hard. Hence `Platform.scenarios()` and `Platform.scenario_list()` are ambigous.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


ixmp/core.py
```diff
@@ -393,6 +393,39 @@ def check_access(self, user, models, access='view'):
         else:
             return {model: result.get(model) == 1 for model in models_list}
 
+    def get_meta(self, *args, **kwargs):
```
zikolach, line 396: we can avoid this redundant calls by adding `get_meta` and `set_meta` to `_backend_direct` list

zikolach: see this commit https://github.com/iiasa/ixmp/pull/353/commits/121a876b7dafd5839c9b2a2ad53e40ff22754482

ixmp/backend/jdbc.py
```diff
@@ -875,15 +881,32 @@ def item_delete_elements(self, s, type, name, keys):
...
+
+        meta = self.jobj.getMeta(model, scenario, version)
+        return {entry.getKey(): _unwrap(entry.getValue())
+                for entry in meta.entrySet()}
+
+    def set_meta(self, meta: dict, model: str = None, scenario: str = None,
+                 version = None):
+        if not (model or scenario or version):
+            msg = ('At least one parameter has to be provided out of: '
+                   'model, scenario, version')
```
zikolach, line 895: I don't think passing just version makes much sense. It should be one of:
- model
- scenario
- model + scenario
- model + scenario + version

ixmp/tests/core/test_meta.py
```diff
@@ -0,0 +1,67 @@
+"""Test meta functionality of ixmp.Platform."""
+
+import pytest
+
+import ixmp
+from ixmp.testing import models
+
+
+sample_meta = {'sample_string': 3, 'another_string': 'string_value'}
```
zikolach, line 9: first entry has not a string value. What is the idea? Maybe just called by mistake

ixmp/backend/base.py
```diff
@@ -902,8 +1020,8 @@ def get_meta(self, s: Scenario):
         """
 
     @abstractmethod
-    def set_meta(self, s: Scenario, name_or_dict, value=None):
-        """Set single or multiple meta entries.
+    def set_scenario_meta(self, s: Scenario, name_or_dict, value=None):
```
zikolach, line 1023: o be duplicated after renaming

zikolach: with that one https://github.com/iiasa/ixmp/pull/353/files#diff-313b0713bfcf3262902e8b94639dd49eR986

khaeru:
I'll wait to review until the tests pass and conflicts resolved.

In the meantime, some questions:
- In addition to worsening (with `add_scenario()`) the *existing* overlap between the semantics of scenario names and `Scenario` objects (as the description notes), this PR creates a new overlap between model names (`add_model()`) and the existing `Model` class. Can this be avoided, e.g. perhaps via the approach described in #274?
- What is the intended behaviour if a user calls `add_timeseries` with model name(s) or scenario name(s) that have not previously been declared with `add_model` or `add_scenario`?
- What is the intended behaviour if a user instantiates `ixmp.Scenario(..., version='new')` with a model name or scenario name that have not been previously declared?

zikolach:
> What is the intended behaviour if a user instantiates ixmp.Scenario(..., version='new') with a model name or scenario name that have not been previously declared?

This behavior did not change - it will create (register) new model/scenario names. Methods were added to allow to create entities which rely on specific model/scenario names before scenario version (run) is created.

zikolach:
> What is the intended behaviour if a user calls add_timeseries with model name(s) or scenario name(s) that have not previously been declared with add_model or add_scenario?

Not sure how does it relate to add_timeseries. Can you please point on the code example or explain what you mean?

zikolach:
> Can this be avoided, e.g. perhaps via the approach described in #274?

I think it makes sense to apply this paradigm of codelists in longer-term perspective. Add mode/scenario names methods are implemented in line with current (probably not efficient) aproach. So later on we can change all such occurrences as adding regions, timeslices, units and new model/scennario names to more generic API in one go.

fonfon:
This PR is now ready to be reviewed.
While we need the meta functionality for the AR6 model meta import, the model/scenario Platform methods aren't that important/urgent.

ixmp/backend/base.py
```diff
@@ -53,6 +54,42 @@ def get_auth(self, user, models, kind):
         """
         return {model: True for model in models}
 
+    @abstractmethod
+    def add_model(self, name: str):
+        """ Add (register) new model name.
```
khaeru, line 59: Please format these correctly:

- No space after the opening triple quote.
- The format for parameters is:
  ```
  name : type
      Description intended four spaces, capitalized, ending with a period.
  ```

See the other methods in this file, and the [numpydoc docstring format](https://numpydoc.readthedocs.io/en/latest/format.html), which is linked from our [style guide](https://docs.messageix.org/en/stable/contributing.html#code-style). Please be sure to compile the docs locally to check that the output is correctly formatted.

ixmp/core.py
```diff
@@ -61,6 +61,13 @@ class Platform:
         'close_db',
         'get_doc',
         'set_doc',
+        'get_meta',
```
khaeru, line 64: These should be sorted in alphabetical order.

khaeru:
General responses to @zikolach's comments:
> This behavior did not change - it will create (register) new model/scenario names. Methods were added to allow to create entities which rely on specific model/scenario names before scenario version (run) is created.

The issue I was trying to highlight was that the existing behaviour was never clearly explained or documented; it was only implicit. Namely, it's not explained to the user anywhere that:
- Model and scenario names are not simply attributes of TimeSeries/Scenario objects, but *distinct entities* in the data model, to which things (e.g. annotations/meta) can be attached.
- These entities were automatically created when new values were used; now they can be created directly.

> I think it makes sense to apply this paradigm of codelists in longer-term perspective. […] So later on we can change all such occurrences as adding regions, timeslices, units and new model/scennario names to more generic API in one go.

Totally agree with this. I think it will also be much easier to explain to the user via documentation, etc.: "ixmp tracks lists of strings or codes that are used for different purposes, and allows attaching meta information to these codes."

In any case, let's merge this PR once the above changes are made, then clean up the Backend API on the Python side in a separate PR, soon. The Java implementation could be simplified later, when time allows.

ixmp/backend/base.py
zikolach: it looks it @fonfon updated it 84ffa4aa139db5cb0da8d306ad841a06b57432de

zikolach:
@khaeru seems like all the requested changes are done. Could you please re-review/approve this PR?

ixmp/backend/base.py
fonfon: The change was done after the comment of Paul.
I noticed two more minor leftovers that I'm cleaning up now (release notes and in tests) so I'll add one more commit.

khaeru:
> seems like all the requested changes are done. Could you please re-review/approve this PR?

Yep! I did see the review re-request from @fonfon a few minutes before your comment—also three commits since, but I guess those are the final ones. I will review as soon as I have time.

PR #399, opened 2021-02-19
Add "quiet" option for GAMSModel

khaeru:
This option is then used to suppress output when e.g. DantzigModel is solved as part of tests where the GAMS output is not of interest. This shortens the pytest output.

Also:
- Mark the `utils.logger()` function as deprecated, remove the few remaining uses of it, and raise a warning when it is called.

## How to review

No review.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -4,6 +4,8 @@ Next release
 All changes
 -----------
 
+- Deprecate :func:`.utils.logger` (:pull:`399`).
+- Add a `quiet` option to :meth:`.GAMSModel.solve` and use in testing (:pull:`399`).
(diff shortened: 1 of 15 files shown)


khaeru:
Comparing [this log](https://github.com/iiasa/ixmp/runs/1945488857) to [this one from the previous PR](https://github.com/iiasa/ixmp/runs/1937329260), the pytest output with `-rA --verbose` is shortened from about 16,000 to 9500 lines, or about 40% by hiding GAMS output during text fixture setup/teardown.

There can still be some improvement, e.g. by turning off the echoing of log output to stderr. But we'll save that for another PR.

PR #421, opened 2021-07-05
Only reset_index() in format_scenario_list() when there are >0 results

LauWien:
With the upgrade of pandas to 1.3.0 `format_scenario_list()` should only `reset_index()` when there are >0 results. 
To satisfy the workflow Lint the package `types-setuptools` needed to be installed.

## How to review

## PR checklist

- [x] Update release notes. ✅

### DIFF
--- a/ixmp/utils.py
+++ b/ixmp/utils.py
@@ -390,10 +390,11 @@ def format_scenario_list(
         platform.scenario_list(model=model, scen=scenario, default=default_only)
         .groupby(["model", "scenario"])
         .apply(describe)
-        .reset_index()
     )
 
-    if not len(info):
+    if len(info):
+        info = info.reset_index()
+    else:
         # No results; re-create a minimal empty data frame
         info = pd.DataFrame([], columns=["model", "scenario", "default", "N"])
 
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -48,5 +48,5 @@ jobs:
       run: |
         pip install mypy
         # Also install packages that contain type stubs
-        pip install pytest genno xarray
+        pip install pytest genno xarray types-setuptools
         mypy .


khaeru (changes requested):
Looks great! Good job on a correct response to the linter error.

The PR template includes this item, which is also needed for this PR:

- [ ] Update release notes.

As it suggests:

> To do this, add a single line at the TOP of the “Next release” section of RELEASE_NOTES.rst, where '999' is the GitHub pull request number:
>
> ```
> - :pull:`999`: Title or single-sentence description from above.
> ```
>
> Commit with a message like “Add #999 to release notes”

I would use the description:
```
- :pull:`421`: Adjust :func:`.format_scenario_list` for changes in :mod:`pandas` 1.3.0.
```

For these notes, it's not necessary to give complete technical detail, just a high-level summary. Readers can click the link back to this PR if they are curious.

LauWien:
Thanks! @phackstock supported.
Next PR I will add the correct commit message for the update of the release notes!

khaeru (approved):
The commit with the release notes update was accidentally pushed directly to `main`: 041d978

ixmp/utils.py
```diff
@@ -390,10 +390,11 @@ def describe(df):
         platform.scenario_list(model=model, scen=scenario, default=default_only)
         .groupby(["model", "scenario"])
         .apply(describe)
-        .reset_index()
     )
 
-    if not len(info):
+    if len(info):
```
phackstock, line 395: I realize I'm a little late to the party and it doesn't really matter but just as a side note, you don't need the `if len(info):` just `if info:` is fine since and empty list evaluates to false (https://docs.python.org/3/library/stdtypes.html#truth-value-testing).

LauWien: True - thanks :)

khaeru: `info` is `pd.DataFrame`, not `list`:
```py
In [1]: df = pd.DataFrame()

In [2]: if df:
   ...:     print("foo")
   ...: 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-559d78b99c9e> in <module>
----> 1 if df:
      2     print("foo")
      3 

~/.local/lib/python3.9/site-packages/pandas/core/generic.py in __nonzero__(self)
   1440     @final
   1441     def __nonzero__(self):
-> 1442         raise ValueError(
   1443             f"The truth value of a {type(self).__name__} is ambiguous. "
   1444             "Use a.empty, a.bool(), a.item(), a.any() or a.all()."

ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

```

phackstock: A right, good point, ignore the comment then

Issue #442, opened 2022-04-19
Cannot commit after add_timeseries() call


Jihoon:
I get an unhandled Java exception during the reporting process of a certain scenario. This error might be related to this [issue ](https://github.com/iiasa/ixmp/issues/324) resolved in the past. 

----

#### Code sample or context

The error is coming directly from this line:
https://github.com/iiasa/message_data/blob/f813f584334417986bc4eb8c2555483ea44b0535/message_data/tools/post_processing/iamc_report_hackathon.py#L392

To replicate that, try this;
```python
import ixmp
import message_ix

mp_ENE = ixmp.Platform("ixmp_dev")
sc_debug = message_ix.Scenario(mp_ENE, "MESSAGEix-GLOBIOM 1.1-M-R12-NGFS", "o_1p5C_NPi2020_600_step1", version=13)

def combined_report(scenario, key, **kwargs):

    '''This combines variant-specific reporting with the standard reporting.
    key: M/BM/BMT to turn on/off corresponding reporting functions
    '''

    from message_data.reporting.materials.reporting import report
    from message_data.tools.post_processing.iamc_report_hackathon import report as reporting

    # Remove existing timeseries and add material timeseries
    print("Reporting material-specific variables")
    print("Now reporting", scenario.model, "/", scenario.scenario, "/", scenario.version)

    if 'M' in key:
        report(scenario, "False")

    # Add timeseries from buildings module
    if 'B' in key:
        add_building_ts(scenario)

    if "legacy" in kwargs:
        legacy_args.update(**kwargs["legacy"])

        print("legacy_args", legacy_args) 
        return reporting(
            mp=scenario.platform,
            scen=scenario,
            **legacy_args,
        )

legacy_args = dict(ref_sol=False, merge_ts=True, run_config="ngfs_run_config.yaml")
combined_report(sc_debug, key = 'M', legacy = legacy_args)
```
#### Expected result

For other usual scenarios, it spits out one xlsx output file without the error.

#### Problem description
Instead, it stops with the following error message.
```
    region                                variable   unit       2025  ...       2080       2090       2100       2110
0  R12_AFR                     Resource|Extraction  EJ/yr  13.521929  ...  14.861251  10.521391  12.236355  15.227066
1  R12_AFR                Resource|Extraction|Coal  EJ/yr   2.436015  ...   0.317465   0.159189   0.070001   0.130232
[...]

PR #443, opened 2022-04-20
Handle ±∞ in TimeSeries.add_timeseries()

khaeru:
#442 appears to be related to handling ±`numpy.inf` in input to `TimeSeries.add_timeseries()`—possibly on Oracle databases.

This PR adds test and fixes the issue by:
- Store the DB properties when the JDBCBackend is first created.
- Check the driver and for infinite values in input data.
- Raise an informative exception.

## How to review

- View test_set_data_inf() in the diff.
- Try calling `.add_timeseries()` with infinite value when connected to a Oracle database via JDBCBackend gives the same kind of exception.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/ixmp/tests/backend/test_jdbc.py
+++ b/ixmp/tests/backend/test_jdbc.py
@@ -10,7 +10,7 @@ from pytest import raises
 
 import ixmp
 import ixmp.backend.jdbc
-from ixmp.backend.jdbc import java
+from ixmp.backend.jdbc import DRIVER_CLASS, java
 from ixmp.testing import DATA, add_random_model_data, bool_param_id, make_dantzig
 from ixmp.testing.resource import memory_usage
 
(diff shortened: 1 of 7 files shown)


khaeru:
@Jihoon since you were able to confirm this when working against an Oracle database, I've asked for your review on the chance you can quickly re-do the steps you used to confirm, except with this branch.

If that's not possible LMK and I will re-request @iiasa/messageix-devs so it can be dispatched to someone else.

ixmp/tests/backend/test_jdbc.py
```diff
@@ -63,44 +64,88 @@ def test_close(test_mp_f, capfd):
 VE = pytest.mark.xfail(raises=ValueError)
 
 
-@pytest.mark.parametrize(
-    "args, kwargs, expected",
-    (
-        # Advertised forms
+class TestJDBCBackend:
+    @pytest.fixture(scope="class")
+    def klass(self):
```
LauWien, line 69: Maybe `jbc` or `jbclass` is more intuitive? But possible just me :)

LauWien (commented):
Looks good to me and works as expected - I like the test class :) Just one suggestion inline.

LauWien (approved):

ixmp/tests/backend/test_jdbc.py
khaeru: I was trying to follow the style of the Python standard library and upstream classes which use either [`cls`](https://docs.python.org/3/library/inspect.html#inspect.getmro) or [`klass`](https://docs.python.org/3/library/logging.html#logging.setLoggerClass) for a variable that will be a reference to a class (because `class` is a special keyword that can't be used).

LauWien: Ah, thank you!

Issue #473, opened 2023-04-06
IXMP on Apple Silicon (ARM architecture)


SiddharthJoshi-Git:
During Solve:

ERROR Reported: java.lang.java.lang.UnsatisfiedLinkError 
Error can be due to CPU/ Java architecture mismatch. Have tried both X86 and ARM builds of JAVA, still this error persists. 

----

#### Code sample or context

--- Warning: The GAMS version [41.1.0] differs from the API version [24.8.3].
java.lang.UnsatisfiedLinkError: Can't load library: [/Users/iiasa/opt/anaconda3/envs/env_x86/lib/python3.9/site-packages/ixmp/backend/jdbc/libgmdjni64.dylib](https://file+.vscode-resource.vscode-cdn.net/Users/iiasa/opt/anaconda3/envs/env_x86/lib/python3.9/site-packages/ixmp/backend/jdbc/libgmdjni64.dylib)
	at java.base[/java.lang.ClassLoader.loadLibrary](https://file+.vscode-resource.vscode-cdn.net/java.lang.ClassLoader.loadLibrary)(ClassLoader.java:2393)
	at java.base[/java.lang.Runtime.load0](https://file+.vscode-resource.vscode-cdn.net/java.lang.Runtime.load0)(Runtime.java:755)
	at java.base[/java.lang.System.load](https://file+.vscode-resource.vscode-cdn.net/java.lang.System.load)(System.java:1953)
	at com.gams.api.gmd.<clinit>(gmd.java:202)
	at com.gams.api.GAMSDatabase.createGMDObject(GAMSDatabase.java:342)
	at com.gams.api.GAMSDatabase.initializeAttributes(GAMSDatabase.java:328)
	at com.gams.api.GAMSDatabase.<init>(GAMSDatabase.java:256)
	at com.gams.api.GAMSWorkspace.addDatabase(GAMSWorkspace.java:1024)
	at com.gams.api.GAMSWorkspace.addDatabase(GAMSWorkspace.java:1010)
	at at.ac.iiasa.ixmp.objects.Scenario.makeGamsDB(Scenario.java:1812)
	at at.ac.iiasa.ixmp.objects.MsgScenario.toGDX(MsgScenario.java:1328)
Unexpected exception formatting exception. Falling back to standard exception
Traceback (most recent call last):
  File "MsgScenario.java", line 1328, in at.ac.iiasa.ixmp.objects.MsgScenario.toGDX
Exception: Java Exception

[...]

PR #478, opened 2023-04-21
Update install.rst

glatterf42:
Add a footnote explaining the most crucial observation in and linking to #473. Also update the 'latest Python version' mentioned there. Note that our pyproject.toml only includes Python versions up to 3.10, so I chose that as the 'latest version'.

## How to review

- Build the documentation and look at the installation page.

## PR checklist

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -78,6 +78,8 @@ jobs:
 
     - uses: r-lib/actions/setup-r@v2
       id: setup-r
+      with:
+        r-version: '4.2.3'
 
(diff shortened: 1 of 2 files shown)


khaeru (approved):
Looks great!

Note that these are the ixmp install instructions. We usually direct new users to the message-ix install instructions, and then ixmp gets installed as a dependency through that process—without them needing to read these separately. These are retained for completeness.

So, we should make a parallel PR with the same changes over there—would you be able to do that, or should I?

glatterf42:
I can do that, but I've got to take care of some other things first.

Issue #463, opened 2022-11-02
Adjust for JPype1 v1.4.1


khaeru:
With the release of JPype1 version 1.4.1, CI jobs have started to fail, e.g. [here](https://github.com/iiasa/ixmp/actions/runs/3326347554/jobs/5540383914), with messages like:
```shell
=================================== FAILURES ===================================
________________________ TestCachingBackend.test_del_ts ________________________

self = <ixmp.tests.backend.test_base.TestCachingBackend object at 0x7f06715dc0d0>
test_mp = <ixmp.core.platform.Platform object at 0x7f05ffaf7cd0>

    def test_del_ts(self, test_mp):
        """Test CachingBackend.del_ts()."""
        # Since CachingBackend is an abstract class, test it via JDBCBackend
        backend = test_mp._backend
        cache_size_pre = len(backend._cache)
    
        # Load data, thereby adding to the cache
        s = make_dantzig(test_mp)
        s.par("d")
    
        # Cache size has increased
        assert cache_size_pre + 1 == len(backend._cache)
    
        # Delete the object; associated cache is freed
        del s
    
        # Objects were invalidated/removed from cache
>       assert cache_size_pre == len(backend._cache)
E       AssertionError: assert 0 == 1
E        +  where 1 = len({(139663741246368, 'par', 'd'):            i         j  value unit\n0    seattle  new-york    2.5   km\n1    seattle   chicago    1.7   km\n2    seattle    topeka    1.8   km\n3  san-diego  new-york    2.5   km\n4  san-diego   chicago    1.8   km\n5  san-diego    topeka    1.4   km})
E        +    where {(139663741246368, 'par', 'd'):            i         j  value unit\n0    seattle  new-york    2.5   km\n1    seattle   chicago    1.7   km\n2    seattle    topeka    1.8   km\n3  san-diego  new-york    2.5   km\n4  san-diego   chicago    1.8   km\n5  san-diego    topeka    1.4   km} = <ixmp.backend.jdbc.JDBCBackend object at 0x7f05ffaf6890>._cache

ixmp/tests/backend/test_base.py:148: AssertionError
_________________________________ test_del_ts __________________________________

    def test_del_ts():
        mp = ixmp.Platform(
            backend="jdbc",
            driver="hsqldb",
            url="jdbc:hsqldb:mem:test_del_ts",
        )
    
        # Number of Java objects referenced by the JDBCBackend
        N_obj = len(mp._backend.jindex)
    
[...]

PR #481, opened 2023-05-03
Support Python 3.11

khaeru:
Take advantage of `genno` >= 1.16, in which `sparse` is an optional/extra dependency.

Previously, via genno → sparse → numba → llvmlite, it was not possible to install/test ixmp with the latest Python; for example, Python 3.11 was released 2022-10-24, but numba 0.57 supporting Python 3.11 was not released until 2023-05-03, a lag of ~6 months.

Notes:
- For #463, we forced a downgrade of JPype1 to < 1.4.1. But JPype1 <= 1.4.0 is not available as binary wheels for Python 3.11. This PR _does not_ force this downgrade for Python 3.11—and the tests seem to pass!
- However, an experiment (see [this run](https://github.com/iiasa/ixmp/actions/runs/4934670781/jobs/8820003482)) confirms that the test failures still occur for Python 3.10, so we retain the workaround.

## How to review

Read the diff and note that the CI checks all pass.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -10,7 +10,10 @@ jobs:
   lint:
     uses: iiasa/actions/.github/workflows/lint.yaml@main
     with:
-      # If the "Latest version testable on GitHub Actions" in pytest.yaml
-      # is not the latest 3.x stable version, adjust here to match:
-      python-version: "3.10"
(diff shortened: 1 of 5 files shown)


glatterf42 (approved):
Looks good to me.

PR #483, opened 2023-06-20
Update RTD config

glatterf42:
As noted by Xiaoyang on Slack, it's not intuitive to choose `[tests]` as the extra requirement for building the docs when we also have `[docs]`. This PR replaces the selection hoping it works out of the box. In the same process, it renames the config file to comply with current recommendations.  

## How to review

- Build the documentation and check that it works as usual.

## PR checklist

- [x] ~Continuous integration checks all ✅~
  Except for `macos-latest-py3.7` per iiasa/ixmp#484.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,4 +4,4 @@ python:
   install:
   - method: pip
     path: .
-    extra_requirements: [tests]
+    extra_requirements: [docs]


glatterf42:
As per the instructions for checking the [RTD build locally](https://github.com/iiasa/message_ix/blob/main/doc/README.rst#read-the-docs), the command completed successfully. We might still be missing some packages required for building the docs by only requiring the extra `[docs]`, but I'm not sure how to test that since the PR originates from my fork (so we cannot simply enable builds for this branch on RTD itself).

khaeru:
> We might still be missing some packages required for building the docs by only requiring the extra `[docs]`, but I'm not sure how to test that since the PR originates from my fork (so we cannot simply enable builds for this branch on RTD itself).

RTD did not originally support building docs for PRs, but has recently added this feature: https://docs.readthedocs.io/en/latest/guides/pull-requests.html. We *could* turn this on, but should first consider security implications, which I haven't had time to do. (For example: code examples in the documentation or code in Sphinx conf.py can be run while the Sphinx build occurs. Could a user abuse this to, e.g. attempt to DDoS our or someone else's public-facing servers?)

So some options would be:
1. Cautiously or temporarily (just for this PR) enable that.
2. Re-open the PR with the same commit pushed to a branch in the main repo.

If (2), then see https://github.com/iiasa/ixmp/issues/484#issuecomment-1606883266 and the comment linked from there.

glatterf42:
From these options, I would choose 2 since I don't know how to assess DDoS potential. However, I'm not sure what I can gain from the linked comment: do you suggest to merge a PR with the suggested changes before opening a new one here?

khaeru:
> However, I'm not sure what I can gain from the linked comment: do you suggest to merge a PR with the suggested changes before opening a new one here?

I meant that, if/when you push to a new branch in the main repo and open a new PR, it could be convenient to add 1 more commit adding the 1 suggested line, at the same time.

glatterf42:
### Superseded by #485

PR #488, opened 2023-07-27
Avoid locking on failed operations

khaeru:
The PR adds a new context manager `.utils.discard_on_error()` that can be used to avoid leaving scenarios in a locked state. This can occur when exceptions are raised and JDBCBackend is in use: an unclean exit means the database is left in a state that marks the scenario as "locked" for use by a particular user.

Originally developed for iiasa/message_data#446.

## How to review

- Read the additions to the docs including examples and ensure they are clear.
- Optionally, try a particular code operation known to fail and leave scenario(s) in a locked state.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/ixmp/core/timeseries.py
+++ b/ixmp/core/timeseries.py
@@ -1,5 +1,5 @@
 import logging
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
 from os import PathLike
 from pathlib import Path
 from typing import Any, Dict, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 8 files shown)


ixmp/core/timeseries.py
```diff
@@ -221,10 +235,13 @@ def transact(self, message: str = "", condition: bool = True):
         >>> # Changes to `ts` have been committed
         """
         # TODO implement __enter__ and __exit__ to allow simpler "with ts: …"
+        from ixmp.utils import discard_on_error as doe
+
         if condition:
             maybe_check_out(self)
         try:
-            yield
+            with doe(self) if discard_on_error else nullcontext():
```
gidden, line 243: suggest to simply spell out what `doe` means for future developers (ie not import as), recognizing that the argument has the same name

gidden (changes requested):
LGTM, thanks @khaeru. Small change requested for future developers.

ixmp/core/timeseries.py
khaeru: Used a longer local name and added an explanatory comment.

gidden:
I think the force push means I can't access my review to change to accept?

gidden (approved):
lgtm

gidden:
Ok, was able to do a new review - thanks @khaeru !

PR #502, opened 2023-12-11
Record Python package versions in GDX files

khaeru:
This PR provides the basic functionality for addressing iiasa/message_ix#747.

The GAMSModel class gains an attribute/parameter (`record_version_packages`) and a method (`record_versions()`); the latter is called from within the `run()` method. This adds a set to the Scenario that is to be solved, with entries like:
```
Set ixmp_version(*,*) Versions of Python packages used to generate this model /
'message_ix'.'3-7-1-dev55+g2e0a9c49', 
'ixmp'.'3-7-1-dev67+g6cab755-d20231122'
'notapackage'.'(not installed)' /;
```
(The list of included packages is according to `record_version_packages`, so can be extended by downstream packages, like `message-ix`, that use or subclass GAMSModel.)

## How to review

Read the diff and note that the CI checks all pass.

Test together with a message_ix PR to be opened next.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ See [`doc/README.rst`](doc/README.rst) for further details.
 
 ## License
 
-Copyright © 2017–2023 IIASA Energy, Climate, and Environment (ECE) program
+Copyright © 2017–2024 IIASA Energy, Climate, and Environment (ECE) program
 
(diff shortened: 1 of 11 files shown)


glatterf42 (approved):
Looks good to me, thanks :)

Issue #501, opened 2023-11-16
Support Python 3.12


khaeru:
This requires, at minimum:
- jpype-project/jpype#1147 —a JPype release that supports Python 3.12.

Issue #494, opened 2023-09-19
Track/work around lp:2035206: missing lxml wheel for cp38-macosx


khaeru:
[No PyPI wheel for 4.9.3-cp38-cp38-macosx](https://bugs.launchpad.net/lxml/+bug/2035206).

This causes CI failures for macos-latest-py3.8: on this os/Python version (and not others), PyPI attempts to compile lxml from source, which fails.

A **workaround** is to pin lxml to != 4.9.3 in the "pytest" CI workflow.
A **fix** would need to happen upstream: publishing the missing wheel, or a new version that is not missing this wheel.

khaeru:
This is reportedly resolved as of 2024-01-06 https://bugs.launchpad.net/lxml/+bug/2035206. #504 will confirm and close.

PR #504, opened 2024-01-08
Address GC with JPype ≥1.5.0

### LINKED ISSUE #463: Adjust for JPype1 v1.4.1

khaeru:
Will close #463, #494, #501.

Also:
- Add pytest-xdist to run tests in parallel on 2+ cores, as with other packages in our stack.  This reduces the pytest run time for some jobs to <1 min.

  One strange thing: the failures of the `test_del_ts` tests noted/discussed in #463 *do not occur* when using pytest-xdist; so those tests are all XPASS on Python ≤ 3.10 on all 3 OS.

## How to review

- Ensure the description of the JPype1 version dependency in the release notes is coherent.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -117,7 +117,13 @@ jobs:
       shell: Rscript {0}
 
     - name: Run test suite using pytest
-      run: pytest ixmp -m "not performance" --verbose -rA --cov-report=xml --color=yes
+      run: |
+        pytest ixmp \
+          -m "not performance" \
+          --color=yes -rA --verbose \
+          --cov-report=xml \
+          --numprocesses=auto
+      shell: bash
 
     - name: Upload test coverage to Codecov.io
       uses: codecov/codecov-action@v3
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -28,8 +28,8 @@ jobs:
         - "3.8"  # Earliest version supported by ixmp
         - "3.9"
         - "3.10"
-        - "3.11" # Latest supported by ixmp
-        # - "3.12"  # Pending JPype support; see iiasa/ixmp#501
+        - "3.11"
+        - "3.12"  # Latest supported by ixmp
 
         # commented: force a specific version of pandas, for e.g. pre-release
         # testing
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -104,10 +104,6 @@ jobs:
         # commented: use with "pandas-version" in the matrix, above
         # pip install --upgrade pandas${{ matrix.pandas-version }}
 
-    - name: TEMPORARY Work around iiasa/ixmp#463
-      if: matrix.python-version != '3.11'
-      run: pip install "JPype1 != 1.4.1"
-
     - name: Install R dependencies and tutorial requirements
       run: |
         install.packages(c("remotes", "Rcpp"))
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -92,10 +92,6 @@ jobs:
       run: echo "RETICULATE_PYTHON=$pythonLocation" >> $GITHUB_ENV
       shell: bash
 
-    - name: Work around https://bugs.launchpad.net/lxml/+bug/2035206
-      if: matrix.python-version == '3.8' && matrix.os == 'macos-latest'
-      run: pip install "lxml == 4.9.2"
-
     - name: Install Python package and dependencies
       # [docs] contains [tests], which contains [report,tutorial]
       run: |
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -14,7 +14,8 @@ Code that imports from the old locations will continue to work, but will raise :
 All changes
 -----------
 
-- Support for Python 3.7 is dropped (:pull:`492`).
+- :mod:`ixmp` is tested and compatible with `Python 3.12 <https://www.python.org/downloads/release/python-3120/>`__ (:pull:`504`).
+- Support for Python 3.7 is dropped (:pull:`492`), as it has reached end-of-life.
 - Rename :mod:`ixmp.report` and :mod:`ixmp.util` (:pull:`500`).
 - New reporting operators :func:`.from_url`, :func:`.get_ts`, and :func:`.remove_ts` (:pull:`500`).
 - New CLI command :program:`ixmp platform copy` and :doc:`CLI documentation <cli>` (:pull:`500`).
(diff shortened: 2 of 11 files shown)


glatterf42:
While looking at these logs, I noticed:
```bash
XPASS ixmp/tests/core/test_scenario.py::TestScenario::test_add_par[args2-kwargs2] Length mismatch between keys and values
```
Taking a closer look, I found
```python
key_or_data = ["new-york", "chicago"]
value = [100, 200, 300]

# This creates as many tuples as possible, stopping when key_or_data runs out
zip(key_or_data, value)
# So there was never nan data in pandas, thus not triggering the ValueError we expect for the test
```

glatterf42:
Update: this only seems to work for python > 3.9, thus some tests are failing. I'll think about as elegant a fix as possible.

khaeru:
I would suggest to use [`itertools.zip_longest`](file:///usr/share/doc/python3.11/html/library/itertools.html#itertools.zip_longest) from the standard library, and keep the `if` block as it was.

Note that the construction:
```py
try:
    func()
except ValueError as e:
    raise ValueError("Helpful message about specific condition X") from e
```
…can only be used if we know for certain that *only* condition X will cause func() to raise ValueError. If there are 2 or more ways it could raise the same class of exception, then the except clause will inappropriately swallow/hide conditions Y or Z from the user, and actually give them an incorrect message about what they've done wrong.

glatterf42:
That's a good suggestion. In theory, though, we could also not reraise our own unique exception since zip(..., strict=True) will raise a ValueError already.

glatterf42:
It's supposed to, at least. On my system, however:
```python
>>> zip(key_or_data, value, strict=True)
<zip object at 0x7fbe4169f9c0>
>>> list(zip(key_or_data, value, strict=True))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: zip() argument 2 is longer than argument 1
```

glatterf42 (approved):
This looks good to me now, thanks :)

PR #513, opened 2024-01-15
Advise users to check GAMS log file

### LINKED ISSUE #511: Advise users to check GAMS log file
https://github.com/iiasa/ixmp/blob/e764f63c38d8327f41c7f9c81318552695de06a2/ixmp/model/gams.py#L195-L224

As suggested by @volker-krey, upon GAMS errors, users should not only check the `.lst` file, but also the `.log` file. I'm not entirely sure what the exact name of these files is, but we can most likely add a line here that checks if such a file exists and points the user to it.

glatterf42:
Closes #511.

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just adding to error message
- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -1,8 +1,10 @@
-.. Next release
-.. ============
+Next release
+============
 
-.. All changes
-.. -----------
+All changes
+-----------
+
+- To aid debugging when execution fails, :class:`.GAMSModel` also displays the path to the GAMS log file (:pull:`513`).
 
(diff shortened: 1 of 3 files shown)


khaeru (approved):
Looks good!

I think this does in fact deserve a release notes entry (almost anything user-facing should, except bug fixes). Maybe something like:
```
- To aid debugging when execution fails, :class:`.GAMSModel` also displays the path to the GAMS log file (:pull:`513`).
```

Please add that, then go ahead to merge.

PR #515, opened 2024-01-25
Save minutes by avoiding duplicate test runs

glatterf42:
This PR removes the trigger for CI runs on pushed to main to avoid duplicate test runs. This avoids duplicate runs.

In order for this to work properly, we should make sure that

- [x] main can only be reached via PRs
- [x] PRs need to be up to date with main
- [x] PRs require several tests to be successful before merging

We might also want to 

- [x] expand the number of tests we require to be successful

All of which can be done via the [settings -> branches](https://github.com/iiasa/ixmp/settings/branch_protection_rules/17031148).

For now, we require the py3.12 checks for macos, ubuntu, and windows to be successful.

## How to review

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Just changing how often tests are run.
- ~[ ] Add, expand, or update documentation.~ Just changing how often tests are run.
- ~[ ] Update release notes.~ Just changing how often tests are run.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -1,8 +1,6 @@
 name: Test
 
 on:
-  push:
-    branches: [ main ]
   pull_request:
     branches: [ main ]
   schedule:


khaeru (approved):

khaeru:
Some notes for this and the parallel PRs iiasa/message_ix#785, iiasa/message-ix-models#150, iiasa/message_data#527:
- Others settings that currently differ, which we could consider aligning:
  - Require 1+ approving reviews.
  - RTD build as a required check.
- Specifically related to https://github.com/iiasa/message-ix-models/pull/150#issuecomment-1910088126, it might be possible to:
  - [force actions/setup-python](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input) to install dev builds for [prerelease versions of Python](https://www.python.org/download/pre-releases/); for example `3.13.0-alpha.3` etc.
  - Allow these jobs to "xfail"; that is, possibly fail without impacting the overall check suite.

  In practice, we will always lag the latest Python: for instance, the moment Python 3.13 is released, some of our dependencies will not be installable on that version for various reasons. But such jobs could allow us to automatically monitor compatibility with newer versions of Python and simply notice (by looking at scheduled runs) when that compatibility is achieved.
  
  IMO this is low priority, since this doesn't cost us a lot of time currently.

khaeru:
@glatterf42 should we merge this? Anything else to be done?

glatterf42:
I think it's good to go :)

PR #516, opened 2024-01-25
Implement ixmp4 shim

khaeru:
This PR is intended to aid @glatterf42 in developing model data features in ixmp4. It is not a complete plan or solution for migrating ixmp-based code, workflows, and research to ixmp4; such a plan would involve many additional elements that are not in scope for this PR.

The PR:
- Adds a IXMP4Backend class that allows ixmp.Platform, ixmp.TimeSeries, and ixmp.Scenario (thus all existing code based on ixmp and message_ix) to talk to ixmp4's API. This functions as a **[shim](https://en.wikipedia.org/wiki/Shim_(computing))** between the established and new/developing APIs.
- Makes minimal changes to allow `ixmp.Platform` backed by IXMP4Backend to be created and used.
- Adjusts or parametrizes tests of the ixmp Platform, TimeSeries, and Scenario classes to run against both JDBCBackend and IXMP4Backend, ensuring the latter provides the same behaviour expected of the former per the test suite.
  - This is a precondition for migrating other code (for instance, in message-ix-models or message_data) to, first, use ixmp4 via an IXMP4Backend, and eventually, ixmp4 directly.
- May serve as a basis for a parallel PR in message_ix.

## Notes

- ixmp4 is not published on PyPI for Python 3.8 and 3.9 and can't be installed in those CI jobs. Possible ways to respond:
  - [ ] Separate `ixmp4` from the `tests` dependency, and don't install it on those Pythons.
  - [ ] Adjust parametrized tests to *not* import ixmp4 or run against IXMP4Backend with those Pythons.
  - [ ] Document that IXMP4Backend is only available for Python >= 3.10.

## How to review

**TBD.** This branch/PR may not necessarily be merged itself, but may serve as the basis for another PR that adds the IXMP4Backend.

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/ixmp/backend/ixmp4.py
+++ b/ixmp/backend/ixmp4.py
@@ -19,6 +19,8 @@ class IXMP4Backend(CachingBackend):
 
         # Add an ixmp4.Platform using ixmp4's own configuration code
         # TODO Move this to a test fixture
+        # NB ixmp.tests.conftest.test_sqlite_mp exists, but is not importable (missing
+        #    __init__.py)
         import ixmp4.conf
 
         dsn = "sqlite:///:memory:"
@@ -29,6 +31,15 @@ class IXMP4Backend(CachingBackend):
 
         # Instantiate and store
         self._platform = ixmp4.Platform(name)
+
+    def get_scenarios(self, default, model, scenario):
+        # Current fails with:
+        # sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: run
+        # [SQL: SELECT DISTINCT run.model__id, run.scenario__id, run.version,
+        # run.is_default, run.id
+        # FROM run
+        # WHERE run.is_default = 1 ORDER BY run.id ASC]
+        return self._platform.runs.list()
 
     # The below methods of base.Backend are not yet implemented
     def _ni(self, *args, **kwargs):
@@ -55,7 +66,6 @@ class IXMP4Backend(CachingBackend):
     get_meta = _ni
     get_model_names = _ni
     get_nodes = _ni
-    get_scenarios = _ni
     get_scenario_names = _ni
     get_timeslices = _ni
     get_units = _ni
--- a/ixmp/backend/ixmp4.py
+++ b/ixmp/backend/ixmp4.py
@@ -1,8 +1,34 @@
+from typing import TYPE_CHECKING
+
 from ixmp.backend.base import CachingBackend
+
+if TYPE_CHECKING:
+    import ixmp4
 
 
 class IXMP4Backend(CachingBackend):
     """Backend using :mod:`ixmp4`."""
+
+    _platform: "ixmp4.Platform"
+
+    def __init__(self):
+        import ixmp4
+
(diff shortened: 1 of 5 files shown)


ixmp/backend/ixmp4.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_scenarios(self, default, model, scenario):
+        # Current fails with:
+        # sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: run
+        # [SQL: SELECT DISTINCT run.model__id, run.scenario__id, run.version,
+        # run.is_default, run.id
+        # FROM run
+        # WHERE run.is_default = 1 ORDER BY run.id ASC]
+        return self._platform.runs.list()
+
+    # The below methods of base.Backend are not yet implemented
```
khaeru, line 44: FYI @glatterf42, the idea is that to remove lines like `name = _ni` below as each method gets a concrete implementation, in parallel with tests that exercise it.

ixmp/tests/core/test_platform.py
```diff
@@ -15,21 +16,48 @@
...
+    def test_init1(self, backend, backend_args):
+        # Platform can be instantiated
+        ixmp.Platform(backend=backend, **backend_args)
+
     def test_getattr(self, test_mp):
         """Test __getattr__."""
         with pytest.raises(AttributeError):
             test_mp.not_a_direct_backend_method
 
+    def test_scenario_list(self, mp):
```
khaeru, line 57: Currently fails (e.g. [here](https://github.com/iiasa/ixmp/actions/runs/7656124831/job/20863732036?pr=516#step:12:904)) with:
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: run
[SQL: SELECT DISTINCT run.model__id, run.scenario__id, run.version, run.is_default, run.id 
FROM run 
WHERE run.is_default = 1 ORDER BY run.id ASC]
```

khaeru:
The Python 3.12 job fails with different exceptions than 3.10 and 3.11, for instance [here](https://github.com/iiasa/ixmp/actions/runs/7656124831/job/20863732388?pr=516#step:12:729):
```
ixmp/backend/ixmp4.py:15: in __init__
    import ixmp4
[...]
/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/sqlalchemy/orm/decl_api.py:303: in __get__
    reg[self] = obj = self.fget(cls)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'ixmp4.data.db.unit.model.Unit'>

    @declared_attr.directive
    def columns(cls):
>       return cls.__table__.columns
E       AttributeError: type object 'Unit' has no attribute '__table__'. Did you mean: '__le__'?

/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/ixmp4/data/db/base.py:76: AttributeError
```
I guess this means ixmp4 does not support Python 3.12, although it is installable.

PR #517, opened 2024-01-29
Address GitHub Actions warnings re: node.js 16

khaeru:
GitHub Actions has started warning about deprecation of Node.js 16, for instance [here](https://github.com/iiasa/ixmp/actions/runs/7691414018):

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-python@v4, ts-graphviz/setup-graphviz@v1.2.0, r-lib/actions/setup-r@v2, actions/cache@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

This PR updates 'first-party' actions as suggested:

- actions/checkout v3 → v4
- actions/setup-python v4 → v5

We cannot yet update the third-party actions:

- [pre-commit/action](https://github.com/pre-commit/action): no update or issue.
- r-lib/actions/setup-r: r-lib/actions#792.
- [ts-graphviz/setup-graphviz](https://github.com/ts-graphviz/setup-graphviz): no update or issue.

Other housekeeping:
- Replace `black` with `ruff format` for code style (iiasa/message_ix#772).
- Tweak the "publish" CI workflow:
  - Remove the trigger for pushes to `main` (analogous to #515).
  - Add `concurrency:` setting to prevent 2 conflicting jobs being started on a release (one triggered by the release event, one by the tag creation event).

## How to review

Read the diff and note that the CI checks all pass.

## PR checklist

- [x] Continuous integration checks all ✅
- [x] ~Add or expand tests;~ coverage checks both ✅
- ~Add, expand, or update documentation.~ N/A, CI changes only
- ~Update release notes.~

PR #518, opened 2024-02-06
Bump versions of GHAs

glatterf42:
This PR is a follow-up of #517 and bumps the remaining versions except for pre-commit, which seems unwilling to release a new version at the moment.

## How to review

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Just CI update
- ~[ ] Add, expand, or update documentation.~ Just CI update
- ~[ ] Update release notes.~ Just CI update

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -62,7 +62,7 @@ jobs:
         cache: pip
         cache-dependency-path: "**/pyproject.toml"
 
-    - uses: ts-graphviz/setup-graphviz@v1.2.0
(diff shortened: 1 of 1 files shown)


khaeru:
Seeing pre-commit/action#189 and pre-commit/action#189, it seems like that action is not "in maintenance mode" per the README but in fact not maintained.

At transport-data/tools#14 I copied the steps into the workflow directly, updated actions/cache and it's not a lot longer than using the third-party one. (The step for clearing the cache daily is actually simpler and easier to read.) That's one option for our stack.

khaeru (approved):

PR #519, opened 2024-03-04
Update deprecated pytest hook arg

glatterf42:
Our whole CI is failing (same for several downstream projects) because one of the pytest hooks here is using a deprecated argument: `startdir` has been replaced by `start_path`. This was kindly pointed out in https://github.com/pytest-dev/pytest/issues/12061, though I have to admit even with the [linked deprecation changelog](https://docs.pytest.org/en/stable/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path), this might not have been entirely clear to me. After all, the changelog says the hook is now receiving 'additional arguments', not 'changed arguments'. Either way, this PR includes the fix.

Should we update the release notes since other projects might have seen failures due to the hook arguments being out of date? If so, we haven't received an issue yet.

## How to review

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Just an internal change
- ~[ ] Add, expand, or update documentation.~ Just an internal change
- ~[ ] Update release notes.~ Just an internal change

### DIFF
--- a/ixmp/testing/__init__.py
+++ b/ixmp/testing/__init__.py
@@ -116,7 +116,7 @@ def pytest_sessionstart(session):
     jdbc._GC_AGGRESSIVE = False
 
 
-def pytest_report_header(config, startdir):
+def pytest_report_header(config, start_path):
     """Add the ixmp configuration to the pytest report header."""
     return f"ixmp config: {repr(ixmp_config.values)}"
 


khaeru (approved):
As noted in Slack, this is ready to merge.

The Windows job failures are acceptable: they occur due to an unrelated issue with the third-party codecov/codecov-action that has also been seen in scheduled (non-PR) runs. The test suite passes, but then the coverage upload stalls, in turn stalling the overall job.

Since this issue with pytest >= 8.1 affects scheduled and PR runs downstream (e.g. iiasa/message_ix#784), we can merge and addressing that issue separately.

PR #510, opened 2024-01-15
Make pytest use one process for CI

glatterf42:
Using pytest-xdist is most likely causing flakiness in our test suite because some tests seem to write/read from processes they should not care about, leading to the common error messages listed below. While we ideally want to implement a more sustainable fix soon (e.g., making the [names of the scenarios/processes unique](https://github.com/iiasa/message_ix/pull/765#issuecomment-1851995857)), we don't want to spend a lot of time rerunning flaky tests manually in the meantime. So this PR makes pytest use just one process as a mitigation suggested by @khaeru.

<details>

<summary> Common error messages </summary>

```bash
FAILED ixmp/tests/test_tutorials.py::test_py_transport - nbclient.exceptions.CellExecutionError: An error occurred while executing the following cell:
------------------
# launch the ix modeling platform using the default back end
mp = ixmp.Platform()

# The following lines have the same effect:
# mp = ixmp.Platform(name='local')  # Default local database
# mp = ixmp.Platform(name='default')  # Default database -> 'local'
------------------


---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
File Platform.java:110, in at.ac.iiasa.ixmp.Platform.<init>()

File Platform.java:163, in at.ac.iiasa.ixmp.Platform.initPlatform()

File DbDAO.java:238, in at.ac.iiasa.ixmp.database.DbDAO.create()

File DbDAO.java:318, in at.ac.iiasa.ixmp.database.DbDAO.updateSchemaIfNecessary()

File Flyway.java:910, in org.flywaydb.core.Flyway.migrate()

File Flyway.java:1238, in org.flywaydb.core.Flyway.execute()

File Flyway.java:910, in org.flywaydb.core.Flyway$1.execute()

File Flyway.java:918, in org.flywaydb.core.Flyway$1.execute()

File Flyway.java:78, in org.flywaydb.core.Flyway.access$100()

File Flyway.java:1037, in org.flywaydb.core.Flyway.doValidate()

Exception: Java Exception

The above exception was the direct cause of the following exception:

org.flywaydb.core.api.FlywayException     Traceback (most recent call last)
File ~/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/ixmp/backend/jdbc.py:297, in JDBCBackend.__init__(self, jvmargs, **kwargs)
    296 try:
--> 297     self.jobj = java.Platform("Python", properties)
    298 except java.NoClassDefFoundError as e:  # pragma: no cover

[...]

Issue #489, opened 2023-08-01
Handling flaky tests


glatterf42:
For roughly the past month, I have collected data on flaky CI tests. The initial idea was to mark them as [flaky](https://github.com/box/flaky), but as per [pytest's docs on flaky tests](https://docs.pytest.org/en/stable/explanation/flaky.html), that should never be a long-term solution. Instead, tests should be (randomly) re-ordered, re-written for more atomic assertions, or split up into different groups to find the root cause of the flaky behavior and eliminate it. We will have to see when time permits this. For now, we could mark them as flaky to save us from re-running them manually. 
Here are the flaky tests of this repository I gathered so far:

# Flaky tests

## Auxiliary

Some error message are common and so long that they would render the table even more complex than it already is. They are included here and referenced by their names in the table below.

### Notebook cell timeout reticulate {#notebook-cell-timeout-reticulate}
```python
nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 10 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
# Load reticulate, used to access the Python API from R
library(reticulate)

# Import ixmp and message_ix, just as in Python
ixmp <- import("ixmp")
-------------------
```

### Notebook cell timeout import packages {#notebook-cell-timeout-import-packages}
```python
nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 10 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
# load required packages
import pandas as pd
import ixmp
-------------------
```

### Notebook cell timeout platform {#notebook-cell-timeout-platform}
```python
nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 10 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
# launch the ix modeling platform using the local default database
mp <- ixmp$Platform()
-------------------
```

### Runtime error DB connection {#runtime-error-db-connection}
```python
RuntimeError: unhandled Java exception: 
[...]

PR #521, opened 2024-03-12
Make tests robust & independent

### LINKED ISSUE #512: Address tests that are flaky with pytest-xdist
As mentioned in https://github.com/iiasa/ixmp/pull/510, pytest-xdist allows us to run tests on multiple tasks simultaneously, decreasing our run time. However, some tests show flaky behaviour since they attempt to read/write from processes they don't belong to. https://github.com/iiasa/message_ix/pull/765#issuecomment-1851995857 mentions a method with which to resolve this flakiness and successfully implements it on several tests in that repo. We should add the same/similar fixes here and then revert the mitigation added in https://github.com/iiasa/ixmp/pull/510.

glatterf42:
Closes #512. 

Analogous to https://github.com/iiasa/message_ix/pull/784, this PR aims at making the ixmp tests less flaky. For now, this is largely done by giving `Scenario`s and `Platform`s test-specific names.

This PR reverts the mitigation from #510. If all goes surprisingly well, this won't be a problem. Either way, it should not be a problem in the end if the intended changes are complete.

## How to review

- Read the diff and note that the CI checks all pass, ideally without manual re-runs.

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just CI tests
- ~[ ] Update release notes.~ Just CI tests


glatterf42:
Some notes here:

- [ ] I've grouped the tutorial tests so that they run on the same worker. This seems like a cheat considering we want to use xdist, but the `*_scenario` tutorials seem to rely on the `*` tutorials running first. However, this grouping might not solve the issue of these required tutorials failing in the first place.
- [ ] I don't understand why `test_export_timeseries_data` is failing. It almost seems like occasionally, `test_mp.export_timeseries_data()` doesn't find the data it's supposed to export. Is this test relying on other tests filling the fixture `test_mp` with data?
- [ ] We should still search the tests for `FLAKY` markers and see if we can resolve more of them.

glatterf42:
The fixes in `test_integration` seem to work locally, but I'm not sure they are optimal yet. They seem to hardcode how many rows in the `scenario` column of the expected dfs need to be changed. I'm not sure we want to do that, but maybe it's okay to expect these particular tests to yield the same results.

glatterf42:
On my system and on the windows-py3.12 platform here, there is this test failure in `test_base.py`: 
```python
FAILED ixmp/tests/backend/test_base.py::TestCachingBackend::test_del_ts - AssertionError: assert 0 == 1
E       AssertionError: assert 0 == 1
E        +  where 1 = len({(133743699349232, 'par', 'd'):            i         j  value unit\n0    seattle  new-york    2.5   km\n1    seattle   c...a    1.8   km\n3  san-diego  new-york    2.5   km\n4  san-diego   chicago    1.8   km\n5  san-diego    topeka    1.4   km})
E        +    where {(133743699349232, 'par', 'd'):            i         j  value unit\n0    seattle  new-york    2.5   km\n1    seattle   c...a    1.8   km\n3  san-diego  new-york    2.5   km\n4  san-diego   chicago    1.8   km\n5  san-diego    topeka    1.4   km} = <ixmp.backend.jdbc.JDBCBackend object at 0x79a3a1141520>._cache

ixmp/tests/backend/test_base.py:161: AssertionError
```
In other words, [these lines](https://github.com/glatterf42/ixmp/blob/main/ixmp/tests/backend/test_base.py#L151-L158) don't seem to garbage collect the cached value correctly. This seems related to #463 and #504, and the issue is resolved by running `s.__del__()` for all python versions. @khaeru, if there's no specific reason to use `del s`, I'd just switch to `s.__del__()`.

khaeru:
> @khaeru, if there's no specific reason to use `del s`, I'd just switch to `s.__del__()`.

We can answer this by first thinking about the intended behaviour of the subject/tested code:
- At the end of a scope where a Scenario object (like `s`) is used, its reference count should decrease, maybe to zero.
- This should result in Python's internal garbage collector choosing to call `Scenario.__del__` (https://docs.python.org/3/reference/datamodel.html#object.__del__ —see the "Note:" in the grey box.)
- That, in turn will free up memory in the Java/JDBC backend.
- In other words, it's not intended that the user should call *either* `del s` or `s.__del__()` manually; rather, whenever they give up all references to the Scenario, the clean-up will happen.

The reason `del s` appears in the test function is to trigger Python's GC earlier, *before* the end of the test function, so that the following assertion can be made about the post clean-up state.

The comment points out that in certain cases, the intended behaviour doesn't work (we haven't diagnosed why). So `s.__del__()` forces the GC to occur and the test to pass. But because it is forced, the test is no longer checking that it will happen autonomously. We might expect that *garbage collection does not work at all* for a user with that combination of JPype and Python.

So to extend: if `s.__del__()` were called in the test in every case, then we no longer will test that Python's built-in reference-counting/GC can or will trigger `Scenario.__del__`—including with the up-to-date Python, JPype, and pandas you mention—and can't claim that this feature is working.

glatterf42:
Thanks for the explanation, so I'll revert/drop that commit and try to figure out how `del s` can call `Scenario.__del__` again.

One other note: I've searched for other `flaky` markers and think I resolved another one, but there's one for [test_jvm_warn](https://github.com/glatterf42/ixmp/blob/bf5d645d8ca82a599d17a52c45a75aa0d08b882e/ixmp/tests/backend/test_jdbc.py#L23-L29) where I don't know what to do. The original flaky error is described in https://github.com/iiasa/ixmp/issues/489 and it just seems that this test sometimes occurs simultaneously as another test reading a big file, which results in a warning. Maybe we can suppress this specific warning, but I'm not sure what else we could do.

glatterf42:
I'm not sure I completely agree with your analysis. The flaky function is called `test_del_ts` and its docstring is `"""Test CachingBackend.del_ts().""". However, we don't explicitly call `CachingBackend.del_ts()` in the test at the moment. 
`del_ts` is a guaranteed part of [`Scenario.__del__`](https://github.com/glatterf42/ixmp/blob/791a87c50556bbfe921c3cab495a80526d1dec38/ixmp/core/timeseries.py#L110-L115), but as the Note box in the docs you linked above points out: 

> `del x` doesn’t directly call `x.__del__()` — the former decrements the reference count for `x` by one, and the latter is only called when `x`’s reference count reaches zero.

In particular, the docs also say:

> CPython implementation detail: It is possible for a reference cycle to prevent the reference count of an object from going to zero. In this case, the cycle will be later detected and deleted by the [cyclic garbage collector](https://docs.python.org/3/glossary.html#term-garbage-collection). A common cause of reference cycles is when an exception has been caught in a local variable. The frame’s locals then reference the exception, which references its own traceback, which references the locals of all frames caught in the traceback.

> Due to the precarious circumstances under which [__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__) methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead.

So the current flaky behaviour we see only reflects exactly that: `del s` doesn't always call `Scenario.__del__`. With that, I'm wondering: should it? And do we want to test it here? 
Especially to the latter, the answer seems to be "No" to me. With this test, according to its name and docstring, we only want to ensure that `CachingBackend.del_ts` works as intended, which it does (locally at least, the flakiness is resolved by always calling `backend.del_ts(s)` explicitly).

I realize that this theoretically leaves part of our code untested, the part where we call `Scenario.__del__`. In practice, however, I expect this finalizer to be called multiple times, e.g. as part of this very function (just for a limited number of python versions). 

In addition, the note that "all exceptions during the execution of __del__() methods are ignored" makes me wonder if we even need the try-clause in our `Scenario.__del__`. Apparently, these exceptions would only result in warnings anyway. Or was it a conscious choice to suppress these warnings?

`

glatterf42:
I'm not sure why `build` couldn't be installed for macos-py3.8, but I don't think that's on us. The error in windows-py3.12 is indeed still flaky and appears to be almost exactly the same as the one above: the test calls `del s` for a number of `Scenario`s, but maybe calling `del_ts` explicitly would be more accurate.

glatterf42:
The tests seem to be doing okay this time around, but I'm not sure we want to keep the solution. In essence, the problem was that the `JDBCBackend` has a `.jindex` attribute, which is a `WeakKeyDictionary`. This means that keys will be discarded automatically once no strong reference to the keys exists at runtime. But when trying to simply call `backend.del_ts()` in the `test_jdbc::test_del_ts`, this didn't always work. My guess is that just as before, a strong reference to some key could be stored as part of a reference cycle. That way, `del s` will not lead to `s.__del__()` being called (which would call `backend.del_ts()`), but calling `backend.del_ts()` is still not enough since we rely on all strong references vanishing to adapt `backend.jindex` and we check that `len(backend.jindex)` changes in our tests. 
So I've now adapted `backend.del_ts` to remove the `ts` from `jindex` after garbage collection to make sure it vanishes, even if a strong reference still exists in some reference cycle. However, this would of course remove `ts` from `jindex` irrespective of the reason a strong reference still exists.

khaeru:
> However, this would of course remove `ts` from `jindex` irrespective of the reason a strong reference still exists.

Hm—in theory, backend.del_ts is not going to get called unless (a) there are no strong references anywhere, so TimeSeries.\_\_del__ is triggered, or (b) the user calls it explicitly.

(But since .jindex is WeakKeyDictionary, the presence (or absence) of `ts` as a key in that should not be making any difference, right? Or maybe I have not fully thought through the implications of keys vs. values vs. both being weak in such collections. If the *key* has some dangling strong ref, then that prevents its removal from the dict and thus the *value* (the JPype/Java/ixmp_source object) also continues to sit there.)

In any case I think this is a safe change to make. Unless the user does (b), this is not going to throw away objects they are still using. I would put the added line in the reverse order, though:
```python
self.jindex.pop(ts, None)
self.gc()
```

khaeru:
Working on #524, I noticed that the Jupyter notebook tests of macos-latest-py3.12 failed several times before eventually passing. They also failed on [the first run on `main`](https://github.com/iiasa/ixmp/actions/runs/8469679494/job/23205559847#step:11:4666) after merging the PR.

glatterf42:
This PR was superseded (and the work started here completed) in #543.

Issue #467, opened 2022-12-05
Address failing tests of R/Jupyter notebooks


khaeru:
Tests of both R tutorial notebooks fail with nbclient.exceptions.CellTimeoutError, e.g. [here](https://github.com/iiasa/ixmp/actions/runs/3599025742/jobs/6069575193#step:14:1676).

This occurs on all jobs that run on ubuntu-latest GitHub Actions runners, but not on Windows or macOS.

Possible responses include:
- Adjust the timeout time.
- Temporarily or permanently skip or xfail the tests.

PR #543, opened 2024-09-11
Make tests robust and independent

glatterf42:
This is a continuation from #521 that puts the branch in the upstream repo rather than my fork to enable windows test runners to access the `GAMS_LICENSE` secret. Please see the original PR for a detailed description, to-dos, etc.

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just CI tests.
- ~[ ] Update release notes.~ Just CI tests.


glatterf42:
I'm just noticing: the windows tests consume much more time than all others. This is not specific to this PR; it was true also for recent dependabot PRs. Not sure if that is worth investigating, definitely not high in priority, I'm afraid.

glatterf42:
I'm rerunning these tests a few times now to see if any flakiness comes up. If it doesn't, I'm happy with the current state of the PR. Just to be sure: my current solution for the flakiness of the tutorial tests is to run these tests on the same worker, which might run counter to our desire to distribute the runs using `pytest-xdist`. However, the tutorial suite in ixmp is small enough so that this doesn't worry me. Most test on recent Python versions still finish in ~3 minutes, which I think is fine.

glatterf42:
The second rerun already revealed something interesting: On windows-py3.10, both of the first tutorial tests timed out when trying to reach the `platform`, but the second ones where happy to start. Unfortunately, for the R tutorials, the first test is required as the second tutorial requires the scenario from the first tutorial to be available. I have now introduced the `pytest-dependency` plugin to skip the second tutorial if the first one failed, but this still doesn't explain why the cells couldn't connect to the platform in the first place.

glatterf42:
https://github.com/iiasa/ixmp/issues/467 is also still open, despite there being a workaround in our code base: the tutorial tests are skipped on ubuntu because they're taking too long. Not sure if we need to keep this issue open, I'm not working on running the tutorials on ubuntu with this PR. We could still use it as a tracking issue if we want to remove the `skip`s one day.

glatterf42:
We discussed `test_del_ts` extensively in #521. 
I'm now choosing to not dive deep into this or the tutorial-related failures since they may well be due to very specific technological details that are not relevant anymore once we transition to ixmp4. As far as I know, ixmp4 is currently only tests on ubuntu, but doesn't contain a single flaky test (despite having ~90% coverage). So for now, I think our best course of action is to mark these tests as flaky (rerunning them when they are failing) and migrating to ixmp4 as soon as we can.
I'd not take the same approach with message_ix or message-ix-models, which won't be replaced in the future as far as I can tell.

glatterf42:
I have rerun the tests now four times and did not observe any flaky behaviour, not even reruns due to flakiness. One time, the `setup-graphviz` action failed due to the backend server being overloaded, but there's not a lot we can do about that in this PR.
I did notice, though, that consistently this happens: `XPASS ixmp/tests/backend/test_jdbc.py::test_del_ts` - https://github.com/iiasa/ixmp/issues/463
This test got excluded for all Python versions <= 3.10 because of an incompatibility with jpype 1.4.1 and Python 1.5.0, which does not seem to persist with pandas > 2.x. I'll remove the `xfail` mark for this and see what happens.

ixmp/tests/test_tutorials.py
```diff
@@ -7,53 +7,73 @@
 
 from ixmp.testing import get_cell_output, run_notebook
 
+group_base_name = platform.system() + platform.python_version()
+
+GHA = "GITHUB_ACTIONS" in os.environ
```
khaeru, line 12: Typically we'd define this in `ixmp.testing`. If it's only used here for now, fine to leave it; but if it's to be re-used, we can move it there.

ixmp/tests/test_tutorials.py
```diff
@@ -7,53 +7,73 @@
...
+    reason="Flaky; see iiasa/ixmp#543",
 )
 
 
+def default_args():
+    """Default arguments for :func:`.run_notebook."""
+    if GHA:
+        # Use a longer timeout
+        return dict(timeout=30)
+    else:
```
khaeru, line 28: Could add `else: # pragma: no cover` or collapse the whole function to a one-liner.

khaeru: Also, the way this function is used makes it seem like it could be a fixture. No?

khaeru (approved):
Thanks for the thorough effort here.

Some comments inline, and I pushed one commit to satisfy mypy in .testing.data. Also:

- R_transport_scenario.ipynb appears to have added cell metadata that are editor-specific. I'd prefer we not commit these unless they are needed for a specific reason.

Aside from those items, good to go.

PR #551, opened 2024-11-28
Pin fastjsonschema for all pythons

glatterf42:
For now, this is just an attempt to fix today's test failure. fastjsonschema changing from v2.20.0 to 2.21.0 was the only difference I could find to yesterday's successful run, though I'm not convinced this is the culprit. 

It seems like pinning fastjsonschema actually works, but I still don't understand why, unfortunately, as it doesn't show up in the traceback.

## How to review

- Run a specific code snippet or command and check the output.
- View the preview build of the documentation and look at a certain page.
- Ensure that changes/additions are self-documenting, i.e. that another
  developer (someone like the reviewer) will be able to understand what the code
  does in the future.

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -107,6 +107,9 @@ jobs:
         # commented: use with "pandas-version" in the matrix, above
         # pip install --upgrade pandas${{ matrix.pandas-version }}
 
+        # TEMPORARY attempt to pin fastjsonschema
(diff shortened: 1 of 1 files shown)


glatterf42:
All checks are passing without any change. So did today's CI, but only after specifically re-running all jobs, not just the one that encountered the issue. We might be required to do that again in the future. I'll close this PR, then.

PR #552, opened 2024-11-28
Implement ixmp4 shim

khaeru:
This PR is intended to aid @glatterf42 in developing model data features in ixmp4. It is not a complete plan or solution for migrating ixmp-based code, workflows, and research to ixmp4; such a plan would involve many additional elements that are not in scope for this PR.

The PR:
- Adds a IXMP4Backend class that allows ixmp.Platform, ixmp.TimeSeries, and ixmp.Scenario (thus all existing code based on ixmp and message_ix) to talk to ixmp4's API. This functions as a **[shim](https://en.wikipedia.org/wiki/Shim_(computing))** between the established and new/developing APIs.
- Makes minimal changes to allow `ixmp.Platform` backed by IXMP4Backend to be created and used.
- Adjusts or parametrizes tests of the ixmp Platform, TimeSeries, and Scenario classes to run against both JDBCBackend and IXMP4Backend, ensuring the latter provides the same behaviour expected of the former per the test suite.
  - This is a precondition for migrating other code (for instance, in message-ix-models or message_data) to, first, use ixmp4 via an IXMP4Backend, and eventually, ixmp4 directly.
- Serves as a basis for iiasa/message_ix#894 and iiasa/message-ix-models#257.

## Notes

- This PR recreates #516 with a new branch name. Using the same branch name in ixmp and message_ix allows to minimize the changes to CI in message-ix-models.
- ixmp4 is not published on PyPI for Python 3.9 and can't be installed in those CI jobs. Possible ways to respond:
  - [x] Separate `ixmp4` from the `tests` dependency, and don't install it on those Pythons.
  - [x] Adjust parametrized tests to *not* import ixmp4 or run against IXMP4Backend with those Pythons.
  - [x] Document that IXMP4Backend is only available for Python >= 3.10.

## How to review

## PR checklist

- [x] Update release notes.


glatterf42:
Thanks for starting this PR! I have reworked the setup part of the tests so that `test_mp`, `mp`, and `test_mp_f` are now parametrized for `"jdbc"` and `"ixmp4"`. Please take a look and tell me what you think :)
In particular, I think some parts are not as robust as they could be, which mainly comes down to me not knowing how we want them to look. For example, `IXMP4BACKEND.__init__` takes only `**kwargs` for now, though I would like it to take something like `name: str, dsn: str | None = None`, and require `dsn` to be there if the platform `name` is not known to ixmp4. 

Aside from questions like this, I think the tests are looking good: most `[ixmp4]` tests are failing with `NotImplementedError`s, but by adding the missing functions, we should see those disappear gradually. I think I should use separate branches for that and keep this branch for the main test setup.

glatterf42:
I'm not sure if the sqlite dsn `":sqlite:///:memory:"` is sufficient for running multiple tasks on the same machine. We might have to integrate the unique `platform_name` here, too, but at least [sqlite's docs](https://www.sqlite.org/inmemorydb.html) say that each DB located at `":memory:"` is distinct from all others, so it might also work just fine.

khaeru:
> [H]ow we want them to look. For example, `IXMP4BACKEND.__init__` takes only `**kwargs` for now, though I would like it to take something like `name: str, dsn: str | None = None`, and require `dsn` to be there if the platform `name` is not known to ixmp4.

Have a look at JDBCBackend.handle_config(): https://github.com/iiasa/ixmp/blob/58e801ca1e6fe474755890b3548a09fe1e074c87/ixmp/backend/jdbc.py#L341-L397

As it stands, this method is called via Config.add_platform(); either from code or using args from the CLI. The dict returned by handle_config() is then stored in the config (file), and *should* contain only valid keyword arguments for the `Backend.__init__` method of the matching Backend subclass. For instance, JDBCBackend.handle_config() returns a dict that should be valid kwargs for `JDBCBackend.__init__`. So IXMP4Backend.handle_config() could ensure there is a DSN, or supply a default, e.g. based on whether the user selects SQLite or Postgres, or has the required package(s) installed.

This mechanism is pretty old, e.g. it predates mature Python `dataclasses`, which would probably be a cleaner way to write something like this (if we were starting from scratch today).

glatterf42:
Thanks for the suggestion :)
I'm not sure how `handle_config()` can help me in this case, though. I'd like to make the logic handling of e.g. `dsn` more robust, but this is (so far) only needed in one case. Most of the tests in `test_platform` use the `mp` or `test_mp` fixtures, which always call `ixmp_config.add_platform()` for either kind of backend. Only `test_init1` doesn't use it and instead checks that a `platform` can still be instantiated. So it would feel like cheating to `add_platform` manually in that test `if backend == "ixmp4"`. And since the test just calls `platform.__init__()`, it will only read the config, not add new things.

Maybe I'll get a better feeling for what we need or what's helpful by first trying to add some other functions that should already be supported by ixmp4 :)

khaeru:
> Maybe I'll get a better feeling for what we need or what's helpful by first trying to add some other functions that should already be supported by ixmp4 :)

Yes, agreed—I'd say our first target is to get some or more tests passing, and then we can look at tidying up any hacks that are required to get to that point. As long as we mark them clearly (`# FIXME` or in commit messages) we can come back and find those.

.github/workflows/pytest.yaml
```diff
@@ -30,7 +30,8 @@ jobs:
         - "3.12"
         - "3.13"  # Latest supported by ixmp
         gams-version:
-        - "43.4.1"  # First version including a macOS arm64 distribution
+        # - "43.4.1"  # First version including a macOS arm64 distribution
+        - "46.1.0" # Version matching the current gamsapi installation
```
glatterf42, line 34: I'm still not sure how best to handle this. Our `gamsapi` dependency has 45.7.0 as the minimum version, but with that one, the windows tests fail (for some reason I didn't investigate closer). Using 46.1.0 works fine, which was released on February 21, 2024, compared to January 19, 2024 for 45.7.0.

glatterf42:
@khaeru, I can't (meaningfully) request your review here because the PR was opened by you. In principle, this should now be good to go :)

I don't understand what's happening with the codecov checks: they seem to not run any new tests (and I've added unit tests for (almost) all new functions, so the test coverage of the new files should be almost 100%). Also, they seem to count a lot of files in `tests` as not tested.

All other failures are flaky and might need reruns to pass.

.github/workflows/pytest.yaml
glatterf42: Never mind that, in the most recent iteration, all windows runs seem to be failing because they can't find a gams installation.

glatterf42: I've now tried pushing 45.7.0 again to test if this is about installing gams from cache or fresh.

glatterf42: Didn't work. Now using 48.1.0 as the corresponding PyPI package is the first one supporting Python 3.13 (and still supporting 3.9).

glatterf42:
So `windows-latest-py3.13` failed because of some flaky test. Rerunning it now seems to yield nothing because I consistently get lots of errors that gams could not be found. Could it be that on the first run, using a new version of gams means it's installed properly (i.e. where `gamsapi` searches for it on Windows), but when rerunning the test, loading and installing from cache sets up gams in a location that `gamsapi` doesn't check? 
Unfortunately, `gamsapi` is still not open source (not that they have any known plans to change that), so likely the only thing for is would be do adjust the gams installation on Windows, because considering how frequently we run into flaky tests with this test suite, we do rely on being able to rerun tests.

glatterf42 (approved):
While this PR is not ready yet, I'm approving in case you have some great idea and now a fix for the gams-issue right away, in which case this formality should not stop you :)

glatterf42:
I have looked into the CI failures again and still don't understand them, but may have found a solution for the most critical one. 

So what I observed in the runs was that whenever I requested a gams version for `setup-gams` that was more recent than what's stored in the the cache, the CI would generally pass (except for some flakiness), but when rerunning the CI on Windows (or just running from a new commit based solely on the cache), all Windows jobs would fail with [numerous errors at test setup](https://github.com/iiasa/ixmp/actions/runs/14516329016/job/40729227140?pr=552) of the sort `Exception: Could not find a GAMS installation, must manually specify system_directory`. 
Reading the [action script](https://github.com/iiasa/actions/blob/main/setup-gams/script.py), I understood why this happens: we're always checking whether the cached `gams.exe` (if there is one) is more recent than the one that would be downloaded and if the latter one is more recent, we download it and extract the files -- unless we're on Windows, where we execute a PowerShell script that runs the official installer. 
So I was thinking: can we really expect that some cached files just work out of the box? And I decided to move the Windows-PowerShell step out of it's existing `if ... elif ... else` clause into its own `if ...`. 
I didn't adjust the other parts, so when we're running on Windows now without having a cache stored, I see this warning:

```
 [D:\a\_actions\iiasa\actions\enh\ixmp4\setup-gams\gams.exe]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
note:  D:\a\_actions\iiasa\actions\enh\ixmp4\setup-gams\gams.exe may be a plain executable, not an archive
unzip:  cannot find zipfile directory in one of D:\a\_actions\iiasa\actions\enh\ixmp4\setup-gams\gams.exe or
        D:\a\_actions\iiasa\actions\enh\ixmp4\setup-gams\gams.exe.zip, and cannot find D:\a\_actions\iiasa\actions\enh\ixmp4\setup-gams\gams.exe.ZIP, period.
```

We could remedy this by keeping an `elif uname.system == "Windows": pass` in the action.
Running this PowerShell script seems to be quite slow in general, unfortunately. In orders of magnitude, `setup-gams` takes ~10 minutes now, while the actual `pytest` step takes about 1.5 minutes. I'm not sure how we could address that, but this sounds like an issue for the `actions/setup-gams` repository in any case.
With this change, however, the Windows tests can now be rerun, independent of whether there is a cache or not, and they seem to pass reliably :)
I'm happy to create a PR from the branch on `setup-gams`, merge it, and drop the `TEMPORARY` commit here :)

Unfortunately, I still don't know why the Codecov checks are failing this badly. Maybe there's also some kind of caching going on there and coverage was measured when the tests were still failing at setup, and because none of the later commits touched the source code/tests again, we're still seeing that result? 
I'm not sure what to do about this other than trying to delete the measured coverage and hoping that dropping the `TEMPORARY` commit triggers a new, more accurate measurement.

Lastly and curiously, we frequently see flaky test behavior in our nightly CI of a particular sort. This only seems to happen on [windows-latest-py3.13](https://github.com/iiasa/ixmp/actions/runs/14587136672/job/40914599238), but with the most recent changes, I'm seeing this exact same error on [ubuntu-latest-py3.13](https://github.com/iiasa/ixmp/actions/runs/14591887743/job/40928820391?pr=552) for the first time. It's still flaky, rerunning solves the issue, but I'm surprised to see this happen. It indicates that the issue is not Windows-specific and that this PR changes the order in which tests are run so that other OSes might be affected.

khaeru:
> [codecov/patch](https://github.com/iiasa/ixmp/pull/552/checks?check_run_id=41014147658) — 95.1% of diff hit (target 98.9%)
> [codecov/project](https://github.com/iiasa/ixmp/pull/552/checks?check_run_id=41014146352) — 98.0% (-1.0%) compared to 8a762c1

These are both acceptable (better than I hoped) given (a) the amount of new code here and (b) we will have subsequent clean-up PRs before release in which we can focus on bringing coverage back up. So no obstacle to merging, and no need to postpone in order to address.

glatterf42:
Especially with the logging, this does not work as intended. Locally, all tests in `test_ixmp4.py` are passing, but on here, the tests checking the logging output fail. This makes me think that this arises from using pytest-xdist and that somehow a logger is either cleared before its contents are checked or the messages are logged to a wrong logger. I tried to resolve this by [creating one log file for each worker](https://pytest-xdist.readthedocs.io/en/stable/how-to.html#creating-one-log-file-for-each-worker) ([adapted](https://docs.python.org/3/howto/logging-cookbook.html) to just fix the filename in `pytest_configure` since we do have handlers configured on the root logger, disabling the recommended `logging.basicConfig` call).
No luck so far, though. I'll see what else I can come up with; otherwise, we could xfail these tests, so that they still provide coverage, but are not expected to pass.

glatterf42:
I can reproduce these failures locally. As soon as `test_ixmp4.py` is not the first file found by pytest, this error occurs. I.e. running `pytest ixmp/tests/backend/test_ixmp4.py ixmp/tests/backend/test_ixmp4_io.py` works fine, running `pytest ixmp/tests/backend/test_ixmp4_io.py ixmp/tests/backend/test_ixmp4.py` produces the three errors we see in all CI tests (or `pytest . --no-benchmark [-n 2]`, for that matter). Not sure what this means, though.

glatterf42:
I think I found a solution :)
Now to clean up...

glatterf42:
With these fixes, the last thing to try before merging: can we use v45.7.0 for gams and gamsapi? This would be the oldest version allowed by our `pyproject.toml` and would thus ensure maximum compatibility.

PR #553, opened 2024-12-12
Apply workaround for Ubuntu Python 3.12-path-R CI failure

glatterf42:
This PR tries to implement the workaround identified in https://github.com/actions/runner-images/issues/11137 for the CI failure on Ubuntu 24.04 with Python 3.12. 
I'm not sure which part(s) of the suggested workaround are actually required, some seem to evaluate to no-ops (e.g. `libssl-dev is already the newest version (3.0.2-0ubuntu1.18).`). I'll play around with the steps to see if the workaround can be made more concise. 

## How to review

- Read the diff and note that the CI checks all pass (in particular the one on `ubuntu-latest-py3.12`).

## PR checklist

- [x] Update tests; coverage checks both ✅
- ~[ ] Add, expand, or update documentation.~ Just CI.
- ~[ ] Update release notes.~ Just CI.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -123,12 +123,6 @@ jobs:
         )
 
         reticulate::py_config()
-
-        # commented: for debugging
-        # print(reticulate::py_config())
-        # reticulate::py_run_string("import os; print(os.environ)")
-
-        # IRkernel::installspec()
       shell: Rscript {0}
 
     - name: Install R dependencies and tutorial requirements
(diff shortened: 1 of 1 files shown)


glatterf42:
Funny enough, calling `reticulate::py_config()` fixes the error on ubuntu, but causes the other OSs to crash. So now we have a step specific to `ubuntu-latest-py3.12`, which does not even need the `IRkernel::installspec()` (as do the other ubuntu runs).

The remaining test error is just flaky.

measrainsey (approved):
thanks @glatterf42 for the changes and given that the tests are failing because they're flaky, this is good with me

PR #555, opened 2025-01-16
Update code quality configuration

khaeru:
- Bump mypy, ruff versions and update config.
  - We now follow the configuration recommended in python/mypy#13916, such that all files are checked with mypy, even if only a few are edited in a particular commit. This, for instance, avoids the situation that a change to type hints in one file causes typing errors in other files that are not checked.
  - IPython/Jupyter notebooks are now included in ruff formatting.
- Update copyright year.
- [x] Work around actions/checkout#2041.

## How to review

**TBD**

## PR checklist

- ~Add, expand, or update documentation.~ N/A, formatting only
- ~Update release notes.~

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -94,29 +94,13 @@ jobs:
       # [docs] requires [tests] which requires [report,tutorial]
       run: uv pip install .[docs]
 
-    - name: Install R dependencies and tutorial requirements
-      # Workaround for https://github.com/actions/runner-images/issues/11137
-      if: ${{ matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12' }}
-      run: |
-        install.packages(c("remotes", "Rcpp"))
-        remotes::install_cran(
-          c("IRkernel", "reticulate"),
-          dependencies = TRUE,
-          # force = TRUE,
-        )
-
-        reticulate::py_config()
-      shell: Rscript {0}
+    - name: "Install libpng-dev"  # for R 'png', required by reticulate
+      if: startsWith(matrix.os, 'ubuntu-')
+      run: sudo apt install libpng-dev
 
     - name: Install R dependencies and tutorial requirements
-      if: ${{ ! (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.12') }}
       run: |
-        install.packages(c("remotes", "Rcpp"))
-        remotes::install_cran(
-          c("IRkernel", "reticulate"),
-          dependencies = TRUE,
-          # force = TRUE,
-        )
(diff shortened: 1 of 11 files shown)


khaeru:
@glatterf42 I could, optionally, expand this PR with 2 kinds of improvements to the GitHub Actions workflows that I have prototyped in `genno` and elsewhere:

1. [pytest.yaml](https://github.com/khaeru/genno/commit/c4cf8fb259eeac1d99a29f5e06bebb684e3efd66): use `uv` to set up Python and invoke pytest and pre-commit. As advertised, this seems to be very fast. It doesn't require any changes to code or packaging, so it seems to be a no-regret change.

2. [publish.yaml](https://github.com/khaeru/genno/commit/98799f2ad31d172d26e6772f9fee5a2fffba2eaf): use `uv build` and `uv publish` instead of other tools. However, because of upstream issues (astral-sh/uv#9513 → pypa/setuptools#4759) this requires a change in the build-backend, for which I went with `hatchling` as suggested in comments on those issues. This isn't a major change, especially as I *only* used the build backend and did not attempt to use any of the other features that hatch advertises. But I think it begins to collide with our discussion of whether to adopt hatch, pdm, poetry, or something else, where we still don't have a clear "yes" on any option. Also, `uv publish` is marked experimental for the time being. So I would maybe avoid to do that in ixmp and the rest of the stack, at least for now.

Please let me know briefly what you think, and I'll proceed and also make parallel changes to message_ix.

glatterf42:
Thanks for this PR :)

I agree with what you say, 1. sounds good, 2. can likely wait a little. Though I have already thought about migrating our package management to `uv` completely myself, i.e. migrate the `pyproject.toml`, generate a lock file, etc. But this can also wait until we can do 2. at least, I think.

khaeru:
Okay, now in place. Compared to [this run](https://github.com/iiasa/ixmp/actions/runs/12850556187):

- macOS:
  - actions/setup-python: 2:49–5:33 
  - astral-sh/setup-uv: 4:11–5:07
- Linux:
  - actions/setup-python: 2:29–4:30
  - astral-sh/setup-uv: 2:38–4:08
- Windows:
  - actions/setup-python: 13:08–15:58
  - astral-sh/setup-uv: 11:39–12:19

So only clear daylight on Windows, but by a few minutes. However there's a lot of variability in the unrelated steps. For the ubuntu-latest-py3.13 jobs, the times for the setup step + "Install the package and dependencies":
- actions/setup-python: 11 seconds + 53 seconds.
- astral-sh/setup-uv: 2 seconds + 6 seconds.

So that's a clear improvement.

glatterf42 (approved):
Thanks, looks great to me :)
Hopeful that this will stabilize our CI :)

PR #556, opened 2025-01-28
Add ixmp4-backed platform functions

glatterf42:
As suggested in https://github.com/iiasa/ixmp/pull/552#issuecomment-2602447693, this PR adds `IXMP4Backend` functions that are required to pass the `test_platform` tests. This is separate to keep the PR reviews manageable.

As far as passing the test_platform tests go, this work is incomplete. In some places, I'm not sure best to proceed. For example, `test_add_region` and `test_add_region_synonym` are currently failing because the regions-DataFrame returned from the ixmp4 Backend does not include the data the tests currently expect. This is in turn due to the ixmp4 DB model for regions is different from the one in ixmp.
Most notably, ixmp4 requires the `hierarchy` parameter, but doesn't use `parent` or `synonym`. I asked @meksor and @danielhuppmann about this and they said that's on purpose and ixmp doesn't really use these parameters anymore, either. So I think it should be fine to adapt the existing tests, but please let me know what you think, @khaeru.

Similarly, ixmp4 does not seem to store timeslices in a similar way to the JDBC. It looks like I'll have to add this feature first, but I'm not really familiar with the IAMC data storage we currently have implemented. 

Another missing feature is the ability to delete stuff in ixmp4. Both optimization data and items as well as `Run`s as a whole still lack this. 

There are several more open questions marked as `TODO` inline. For example: `ixmp4.Run.clone()` doesn't support `first_model_year` and `platform_dest`. How crucial are these features?

## How to review

**Required:** describe specific things that reviewer(s) must do, in order to ensure that the PR achieves its goal.
If no review is required, write “No review: …” and describe why.

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/ixmp/backend/base.py
+++ b/ixmp/backend/base.py
@@ -204,12 +204,13 @@ class Backend(ABC):
         """
 
     @abstractmethod
-    def get_nodes(self) -> Iterable[tuple[str, Optional[str], str, str]]:
+    def get_nodes(self) -> Iterable[tuple[str, Optional[str], Optional[str], str]]:
         """Iterate over all nodes stored on the Platform.
 
         Yields
         -------
         tuple
+
             The members of each tuple are:
 
             ========= =========== ===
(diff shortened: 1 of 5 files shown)


glatterf42:
Superseded by #557 from within this repo.

PR #560, opened 2025-03-03
Enable `populate_test_platform()` with `IXMP4Backend`

glatterf42:
This PR adds a little more functionality to the `IXMP4Backend`, which is enough to make `populate_test_platform()` pass. In turn, this includes `make_dantzig(solve=True)`, so this feels like quite a significant milestone. Enough, at least, to now turn to message_ix and see if I can't get `westeros_baseline` to run. After all, everything making that more fancy than the transport tutorial should be handled in message_ix, I think.

One item deserves special attention, maybe: the `JDBCBackend` has certain units pre-defined, it seems. For `make_dantzig()`, I've included a check for now: when we're on an `IXMP4Backend`, we quickly add the required units. However, this only covers this one test function and we probably want to come up with a solution that covers all cases. Which likely means:

1. Figure out which units are pre-defined
2. Set up a check in `Platform.__init__()` that defines all these units on an `IXMP4Backend`

## How to review

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/ixmp/backend/ixmp4.py
+++ b/ixmp/backend/ixmp4.py
@@ -1,24 +1,22 @@
 import logging
 from collections.abc import Generator, Iterable, MutableMapping, Sequence
-from typing import TYPE_CHECKING, Any, Literal, Optional, Union, cast
+from typing import Any, Literal, Optional, Union, cast
 
 import pandas as pd
+from ixmp4 import Platform as ixmp4_platform
+from ixmp4.core import Run
+from ixmp4.core.optimization.equation import Equation
+from ixmp4.core.optimization.indexset import IndexSet
+from ixmp4.core.optimization.parameter import Parameter
+from ixmp4.core.optimization.scalar import Scalar
+from ixmp4.core.optimization.table import Table
+from ixmp4.core.optimization.variable import Variable
+from ixmp4.data.backend.base import Backend as ixmp4_backend
 
 from ixmp.backend.base import CachingBackend
 from ixmp.core.platform import Platform
 from ixmp.core.scenario import Scenario
 from ixmp.core.timeseries import TimeSeries
-
-if TYPE_CHECKING:
-    from ixmp4 import Platform as ixmp4_platform
-    from ixmp4.core import Run
-    from ixmp4.core.optimization.equation import Equation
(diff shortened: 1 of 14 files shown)


ixmp/backend/ixmp4.py
```diff
@@ -184,12 +213,16 @@ def clone(
...
-        )
         cloned_run = self.index[s].clone(
             model=model, scenario=scenario, keep_solution=keep_solution
         )
+        # Instantiate same class as the original object
+        cloned_s = s.__class__(
+            platform_dest,
+            model,
+            scenario,
+            version=cloned_run.version,
```
khaeru, line 224: Does this result in the clone having the same version number as the original?

khaeru:
Great progress 🚀 

> The `JDBCBackend` has certain units pre-defined, it seems. For `make_dantzig()`, I've included a check for now: when we're on an `IXMP4Backend`, we quickly add the required units. However, this only covers this one test function and we probably want to come up with a solution that covers all cases. Which likely means:
> 
> 1. Figure out which units are pre-defined
> 2. Set up a check in `Platform.__init__()` that defines all these units on an `IXMP4Backend`

Yes, agreed this is the right approach. We can imagine a migration/deprecation path:

1. For now, silently and always add these to any Platform on an IXMP4Backend.
2. Later, code can notice when users try add data with these unit-strings to an IXMP4Backend-backed platform; make a noisy warning; and then automatically populate the units; but otherwise not do so.
3. Later than (2), the auto-population can be dropped, and it becomes the responsibility of user code (e.g. message-ix-models) to populate the units it intends to use.

ixmp/backend/ixmp4.py
glatterf42: It only means that the cloned `Scenario` object has the same version as the cloned `Run` object. I'm not sure how ixmp4 handles/changes versions upon cloning, but there was some test that failed with the previous version of the code because the version of the cloned `Scenario` object was set incorrectly (not to the version of the cloned `Run`).

glatterf42:
I'll merge these small PRs into one to provide a better overview of what's changing in ixmp to add support for ixmp4.

PR #565, opened 2025-03-25
Enable running the message_ix Westeros Baseline tutorial completely

glatterf42:
This PR adds the remaining changes required to run the whole Westeros Baseline tutorial from message_ix on an `IXMP4Backend`.

Switching to a persistent local ixmp4 DB might require the `alembic.ini` file from ixmp4 to be present at a specific location (I've reached out to @meksor and @danielhuppmann to inquire if that's on purpose). The easiest option to achieve that right now seems to be:

1. `git clone` ixmp4
2. Set up a venv (preferably using poetry) and install ixmp4 there
3. Activate the venv and `ixmp4 platforms add ixmp4-local`

This should register the DB correctly and run all migrations so that it's up-to-date.

A persistent DB is needed because otherwise, the follow-up tutorials will start from an empty in-memory DB again, i.e. not seeing the solved baseline scenario, which is always used as the basis to make a clone from.

## How to review

- Read the diff.
- Run a westeros baseline tutorial in message_ix based on this branch and see that it works :)

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/ixmp/_config.py
+++ b/ixmp/_config.py
@@ -70,7 +70,11 @@ def _platform_default():
         },
         "ixmp4-local": {
             "class": "ixmp4",
-            "dsn": "sqlite:///:memory:",
+            "dsn": (
+                "sqlite:///"
+                f"{Path.home().joinpath('.local', 'share', 'ixmp4', 'databases')}"
+                "/ixmp4-local.sqlite3"
+            ),
         },
     }
 
--- a/ixmp/backend/io.py
+++ b/ixmp/backend/io.py
@@ -430,6 +430,22 @@ def _align_records_and_domain(
         return {key: records[key] for key in domain_order}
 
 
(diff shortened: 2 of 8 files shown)


khaeru:
Great work!

> Switching to a persistent local ixmp4 DB might require…

For this kind of run-once migration task, it's very valuable to have this kind of clear list of steps to be performed. I imagine/hope most of these could be scriptable, so we could capture this in a utility function that's part of the shim, and then decide if/when to run that function for each user—once, as necessary, etc. This could save us a *lot* of work in manual steps/assistance for migrating users/correcting mistakes they make in the manual steps/etc.

khaeru:
However to be clear, that's part of the step of "preparing this code for real users" which follows on "making it function"—I don't mean to suggest an expansion of scope.

glatterf42:
I'll merge these small PRs into one to provide a better overview of what's changing in ixmp to add support for ixmp4.

PR #568, opened 2025-04-25
Adjust dependencies and imports

khaeru:
This PR adjusts how the `ixmp4` dependency is enacted, in order to ease adoption downstream.

1. In pyproject.toml:
   - Previously: `ixmp[tests]` depended conditionally (only for Python 3.10 and greater) on `ixmp[ixmp4]` → `ixmp4`.
   - Now: `ixmp[tests]` → `ixmp[ixmp4]` which depends conditionally on `ixmp4`.
  
   This allows that downstream packages, e.g. message_ix, can depend on `ixmp[tests]` and `ixmp[ixmp4]`, without needing to insert/modify their own restrictions on particular Python versions(s) not supported by ixmp4.
2. Import of `IXMP4Backend` and thus `ixmp4` in order to populate `.backend.BACKENDS`:
   - Previously, this took place in the top-level \_\_init__.py (thus on every `import ixmp`) and conditional only on the Python version. This would not work e.g. in the case where the user installs `message_ix`, `ixmp`, but *not* `ixmp[ixmp4]`.
   - Now, the import happens as-needed, within a new function `.backend.get_class()`. This:
     - parallels a pattern already used by `.model.get_model()`.
     - means that any 'heavy' imports of/via ixmp4, JPype1, etc. only happen as/when necessary.
       Thus a user using solely *either* backend will not be repeatedly importing dependencies of the other.

As well:
- Clean some of the pytest utilities and infrastructure.
- Improve docstrings.

## How to review

- Read the diff.

## PR checklist

- [ ] Update release notes.


ixmp/backend/ixmp4.py
```diff
@@ -68,48 +149,42 @@ class IXMP4Backend(CachingBackend):
...
         self._platform = ixmp4_platform(_backend=self._backend)
 
+        if opts.jdbc_compat:
+            for u in "???", "GWa", "USD/kWa", "cases", "kg", "km":
+                try:
+                    self.set_unit(u, "For compatibility with ixmp.JDBCBackend")
+                except NotUnique:
+                    # Already exists
+                    # NB The exception class is actually UnitNotUnique, but this is
+                    #    created dynamically and is not importable
```
glatterf42, line 181: I'm reasonably sure you could import `Unit` from `ixmp4.core` and then catch `Unit.NotFound`, but not sure if we need to do that. The only `NotFound` raised by `set_unit()` should be the `Unit.NotFound`.

ixmp/backend/ixmp4.py
```diff
@@ -68,48 +149,42 @@ class IXMP4Backend(CachingBackend):
     # calls for in-memory DBs
     backend_index: dict[str, "ixmp4_backend"] = {}
 
-    def __init__(self, _name: Optional[str] = None) -> None:
-        from ixmp4.conf.base import PlatformInfo
+    def __init__(self, **kwargs) -> None:
```
glatterf42, line 152: I would prefer to type hint kwargs, too. This would mean something of the sort (if I understand correctly):

```python

class InitKwargs(TypedDict, total=False):
    ixmp4_name: str
    dsn: str
    jdbc_compat: bool

class IXMP4Backend(CachhingBackend):
    ...
    def __init__(self, **kwargs: Unpack[InitKwargs]) -> None:
        ...
```

When a user starts typing `IXMP4Backend(`, this would then offer the allowed keys as auto-completions, which I find nice :)

ixmp/model/__init__.py
```diff
@@ -1,11 +1,36 @@
...
+
+
+def get_model(name: Optional[str], **model_options) -> "ixmp.model.base.Model":
+    """Return a model instance for `name` (or the default) with `model_options`.
+
+    Note that unlike :func:`.backend.get_class`, this function creates a new instance.
+
+    Parameters
+    ----------
+    name :
```
glatterf42, line 25:
```suggestion
    name : str, optional
```
Why not add the type hint here, too?

glatterf42 (approved):
Thanks, this generally looks good to me, I'd only like to see the `IXMP4Backend.__init__()` kwargs annotated. 

I'm now looking into why the tests on Windows and Macos fail, which seems to be about the path to the DB file.

khaeru:
> I'm now looking into why the tests on Windows and Macos fail, which seems to be about the path to the DB file.

I have an inkling and just ran out of time to confirm—in short ixmp4 seems to hard code the same path on all OSes. There's a couple related changes I'd like to make this morning before merging. 

Thanks also for the review, will respond shortly.

ixmp/model/__init__.py
khaeru: In short, I think it's now redundant:
- Our Sphinx configuration result in the built docs displaying argument types that get derived from the type hints.
- A person reading the code itself can see the hints directly.

ixmp/backend/ixmp4.py
khaeru: It seems like [`Unpack` was only added in Python 3.11](https://docs.python.org/3/library/typing.html#typing.Unpack)—but I can just add explicit kwargs.

It would be nice if there was some shorthand that allowed to say "The same kwargs and with the same types as the `Options` class."

PR #571, opened 2025-05-02
Clone Scenario.scheme on all clone() calls

glatterf42:
Some of the tests on https://github.com/iiasa/message_ix/pull/894 fail because the `.scheme` property of a Scenario is not cloned when calling `clone()` on an IXMP4Backend (and instead set to "IXMP4-MESSAGE"), which message_ix doesn't expect. It expects "MESSAGE", but we can't hardcode that either because some ixmp tests expect something else (and we generally want to avoid hardcoding things, of course). 

So this PR adjusts the `clone()` call on IXMP4Backend to also clone the `.scheme`. 

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just a small internal fix.
- ~[ ] Update release notes.~ Just a small internal fix.

### DIFF
--- a/ixmp/backend/ixmp4.py
+++ b/ixmp/backend/ixmp4.py
@@ -349,6 +349,7 @@ class IXMP4Backend(CachingBackend):
             model,
             scenario,
             version=cloned_run.version,
+            scheme=s.scheme,
         )
         self._index_and_set_attrs(cloned_run, cloned_s)
         return cloned_s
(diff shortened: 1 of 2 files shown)


khaeru (approved):
I think this is the kind of small nuance of ixmp behaviour that previously lacked tests. If it had been tested, we would have caught this in #552. So, added the test that was missing.

Otherwise good to go once CI passes.

PR #572, opened 2025-05-07
Support message_ix test suite

glatterf42:
Trying to run the message_ix test suite on IXMP4Backend shows that most things are already working :)
A few functions where not tested properly here, so only the message tests revealed how they need to be fixed, which this PR does. 

This likely indicates that we _should_ expand our test suite here if we want to ensure this kind of behaviour, but I'm not sure how easily we can do that when e.g. `cat_list()` is only defined on a `message_ix.Scenario`.

## How to review

## PR checklist

- [x] Update tests; coverage checks both ✅
- ~[ ] Add, expand, or update documentation.~ Internal fixes.
- ~[ ] Update release notes.~ Internal fixes.

### DIFF
--- a/ixmp/backend/base.py
+++ b/ixmp/backend/base.py
@@ -649,15 +649,16 @@ class Backend(ABC):
         tuple
             The members of each tuple are:
 
-            ======== ===== ===
-            ID       Type  Description
-            ======== ===== ===
-            region   str   Region name
-            variable str   Variable name
-            unit     str   Unit symbol
-            year     int   Year
-            value    float Data value
-            ======== ===== ===
+            ========= ===== ===
+            ID        Type  Description
+            ========= ===== ===
+            region    str   Region name
+            variable  str   Variable name
+            unit      str   Unit symbol
+            subannual str   Type of subannuality
+            year      int   Year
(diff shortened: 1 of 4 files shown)


glatterf42:
The remaining failure is flaky; not sure what's up with codecov again. There's hardly anything here that changed functionally and everything that did change is tested in the message_ix test suite. I suggest to accept the coverage drop until we have implemented all functions on IXMP4Backend that are going to be implemented, which will let us remove some `@pytest.mark.jdbc`s, in turn increasing coverage. But I could also write up some more tests here for internal functions, if you prefer.

khaeru:
> There's hardly anything here that changed functionally and everything that did change is tested in the message_ix test suite.

I suspect that's the issue: as you noted in the PR description, the coverage is (improperly) only downstream, and not in the ixmp suite itself. Looking at https://app.codecov.io/gh/iiasa/ixmp/pull/572 it seems, for instance, that missing lines are in `IXMP4Backend.cat_get_elements()`; there are no direct tests for this backend method in the ixmp suite.

I suspect we could devise a way to cover these fairly easily—but not as part of this PR. So let's accept the drop for now.

PR #573, opened 2025-05-09
Adjust to ixmp4 changelog and attribute name change

glatterf42:
While I was working on getting the tutorials to run, @meksor implemented the changelog and versioning in ixmp4. This still has to be done for all optimization items, which I will do soon, but not before the 3.11 release. So there will be more adjustment which will likely introduce a proper `commit()` structure in the shim layer. 

For the time being, it's mainly iamc/timeseries data that is versioned. This seems to affect only one function here, `IXMP4Backend.set_data()`, but let's see what the message_ix test suite reveals, too, before merging this PR. 
For that function, I was wondering whether acquiring the lock (or calling `transact()` should not rather be part of the test functions calling this function, but the test functions only interact with `Scenario` objects, which are only under their hood mapped to `Run` objects. That's why I put the `run.transact()` inside `set_data()` for now, please let me know if that's correct.

In addition, the ixmp4 API changed to more clearly reflect what type we're dealing with when looking at e.g. `Parameter.indexsets`: this return a `list[str]` and has thus changed to `Parameter.indexset_names` (in line with `Parameter.column_names`).

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just internal adaptation.
- ~[ ] Update release notes.~ Just internal adaptation.

### DIFF
--- a/ixmp/backend/ixmp4.py
+++ b/ixmp/backend/ixmp4.py
@@ -261,7 +261,7 @@ class IXMP4Backend(CachingBackend):
                 run.scenario.name,
                 # TODO What are we going to use for scheme in ixmp4?
                 "IXMP4Run",
-                run.is_default(),
+                run.is_default,
                 # TODO Change this from being hardcoded
                 False,
                 # TODO Expose the creation, update and lock info in ixmp4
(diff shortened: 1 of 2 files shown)


khaeru (approved):
I see one known flaky test of JDBCBackend, but everything else seems to be passing, so LGTM!

Issue #546, opened 2024-09-26
Make more tests robust and independent


glatterf42:
Unfortunately, despite our recent improvements in #543, we still see some apparent flakiness. Its main source seems to be the setup-graphviz action on the windows runners, which sometimes fails like this:
```bash
 C:\ProgramData\Chocolatey\bin\choco.exe install graphviz
Chocolatey v2.3.0
Installing the following packages:
graphviz
By installing, you accept licenses for the packages.
Downloading package from source 'https://community.chocolatey.org/api/v2/'
[NuGet] Error downloading 'Graphviz.12.1.1 : chocolatey-core.extension [1.3.5, )' from 'https://community.chocolatey.org/api/v2/package/Graphviz/12.1.1'.
[NuGet] Response status code does not indicate success: 503 (Service Unavailable: Back-end server is at capacity).
[NuGet] Error downloading 'Graphviz.12.1.1 : chocolatey-core.extension [1.3.5, )' from 'https://community.chocolatey.org/api/v2/package/Graphviz/12.1.1'.
[NuGet] Response status code does not indicate success: 503 (Service Unavailable: Back-end server is at capacity).
Graphviz not installed. An error occurred during installation:
 Error downloading 'Graphviz.12.1.1 : chocolatey-core.extension [1.3.5, )' from 'https://community.chocolatey.org/api/v2/package/Graphviz/12.1.1'.
Graphviz package files install failed with exit code 1. Performing other installation steps.
The install of Graphviz was NOT successful.
Graphviz not installed. An error occurred during installation:
 Error downloading 'Graphviz.12.1.1 : chocolatey-core.extension [1.3.5, )' from 'https://community.chocolatey.org/api/v2/package/Graphviz/12.1.1'.

Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Failures
 - Graphviz (exited 1) - Graphviz not installed. An error occurred during installation:
 Error downloading 'Graphviz.12.1.1 : chocolatey-core.extension [1.3.5, )' from 'https://community.chocolatey.org/api/v2/package/Graphviz/12.1.1'.
```

In addition, there was a new flaky test today:

```python
      ____________________________ test_connect_message _____________________________
[gw0] win32 -- Python 3.9.13 C:\hostedtoolcache\windows\Python\3.9.13\x64\python.exe

[...]

PR #577, opened 2025-05-25
Adjust test_jdbc.test_del_ts()

khaeru:
1. Adjust a test which (per https://github.com/iiasa/ixmp/issues/546#issuecomment-2482142993) is still flaky after the adjustments of #543:
   - Use non-overlapping scenario names.
   - Loosen expectation on getrefcount(): only log offending entries, since other assertions pass even when these exist.
   - Check id() of del'd object against Id() of remaining jindex keys.
2. Update the install instructions to re-sync with (and reference/avoid duplicating) the corresponding MESSAGEix install instructions.

## How to review

- Read the diff.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -12,6 +12,7 @@ All changes
   - Improve the :program:`ixmp platform add` :doc:`command <cli>` to support adding :class:`.Platform` with :class:`.IXMP4Backend` (:pull:`575`).
 
 - Refine the method of locating the GAMS :attr:`~.GAMSInfo.executable` (:pull:`564`, :issue:`456`, :issue:`523`, :issue:`563`).
+- Update installation instructions to align with current |MESSAGEix| install documentation (:pull:`577`).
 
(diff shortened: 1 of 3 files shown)


glatterf42 (approved):
Thanks for this cleanup, looks great to me :)

Since I see three flaky errors on the latest CI run here and they are all from `test_del_ts`, I'm wondering: could this be because we use the same platform for all these tests? 
The scenario names are already unique, but still `len(backend.jindex)` seems to sometimes pick up additional scenarios that are not supposed to be there. I'm wondering if that is because we hardcode generic platform details in this test; we could try making the platform name unique just like we do with the scenario names. 
Alternatively, we can maybe use the `test_mp_f` fixture for this.

glatterf42:
No, the flakiness persists. We can drop my latest commit, it seems.

khaeru:
> forcibly clear the .jindex at the start of the test.

This appears to have worked, so please merge once the CI completes.

PR #593, opened 2025-08-28
Fix minor code issues 08/25

glatterf42:
This PR does two things:

1. It adjusts our type hints so that our Code Quality check passes again (respecting yesterday's release of pandas-stubs and the latest versions of mypy and ruff).
2. It adds `urllib3==2.5.0` to our `doc/requirements.in`. For some reason, dependabot fails to perform the upgrade (probably because `requests` doesn't require this version of `urllib3` yet, so we could also just wait), but wants to have that, so I included the fix here. 

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just maintenance.
- ~[ ] Update release notes.~ Just maintenance.


glatterf42:
For some reason, these two tests are failing all the time today:

```
 FAILED ixmp/tests/test_tutorials.py::test_py_transport[jdbc]@Darwin3.11.9-0 - nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 30 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
# Launch the ix modeling platform using the default back end
mp = ixmp.Platform()

# The following lines have the same effect:
# mp = ixmp.Platform(name='local')  # Default local database
# mp = ixmp.Platform(name='default')  # Default database -> 'local'
-------------------
FAILED ixmp/tests/test_tutorials.py::test_py_transport_scenario@Darwin3.11.9-0 - nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 30 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
# launch the ix modeling platform using the default local database
mp = ixmp.Platform()
-------------------
```

This is not particular to this PR, it also happens in our nightly CI. I'm not sure why, but I'm hesitant to put time into research this because it's about Jupyter notebook cells timing out, which seems like it's flaky.

ixmp/backend/ixmp4.py
```diff
@@ -721,11 +799,11 @@ def _add_data_to_set(
             # NOTE ixmp_source silently ignores duplicate data; replicate here
             # This could be improved by adding data without loading the indexset first,
             # but this requires users to ensure their data are valid
-            indexset = self.index[s].optimization.indexsets.get(name=name)
+            indexset = run.optimization.indexsets.get(name=name)
             if key not in indexset.data:
-                indexset.add(key)
+                self._backend.optimization.indexsets.add_data(id=indexset.id, data=key)
```
glatterf42, line 804: Using `_backend` functions like this avoids the requirement for a `Run` to be locked. I'm not sure how we want to handle the `lock` requirement (and `Run.transact()` context) more broadly, but I thought we likely want to have something like `Scenario.transact()`, which in turn calls `Run.transact()`, and within that perform operations like `init_item()` or `item_set_elements()`. So if these functions themselves contain a `Run.transact()` statement, I'm not sure if that works (because the `Run` might be locked from the outer context manager still). 
On the flip side, this means that using these functions outside of `Scenario.transact()` blocks means the `Run` is not safe against concurrent usage (or rolled-back on errors).

There is also the option to write some custom code here that just checks if the `Run` is locked. `Run` comes with the `.owns_lock` property and `_lock()` and `_unlock()` functions, though these are marked as private, so we should probably not rely on them.

ixmp/backend/ixmp4.py
```diff
@@ -500,7 +512,8 @@ def clear_solution(self, s: Scenario, from_year: Optional[int] = None) -> None:
                     "s_clear_solution(from_year=...) only valid for ixmp.Scenario; not "
                     "subclasses"
                 )
-        self.index[s].optimization.remove_solution()
+        with self.index[s].transact("Clear solution for ixmp.Scenario"):
```
glatterf42, line 515: In this case, the `_backend` repositories don't have a function like `remove_solution()` (i.e. a function that takes care of the collective removal of solution data from Equations and Variables of a particular Run). Of course, we could copy the logic from ixmp4, but I don't think that is particularly efficient, so it might change soon (and then offer a better way of calling `_backend` functions directly from here) and I wanted to keep the code DRY where possible.

ixmp/backend/ixmp4.py
```diff
@@ -954,25 +1044,30 @@ def item_delete_elements(
...
+            )
 
     def delete_item(
         self, s: Scenario, type: Literal["set", "par", "equ"], name: str
     ) -> None:
         # Locate the item. If `type` maps to >1 IXMP4 model data type, try each repo.
         item = self._find_item(s=s, name=name, types=CLASS_FOR_IX_TYPE[type])
         # Access the repository containing objects of `item`s type; delete
-        self._get_repo(s=s, type=item.__class__).delete(item=name)
+        # with self.index[s].transact("Delete item for ixmp.Scenario"):
```
glatterf42, line 1069: This line should just be deleted, sorry.

ixmp/backend/ixmp4.py
```diff
@@ -1036,23 +1133,30 @@ def cat_set_elements(
...
         # Get or create the 'type_name' indexset and 'cat_name' table
         try:
-            category_indexset = self.index[ms].optimization.indexsets.get(
-                name=f"type_{name}"
-            )
+            category_indexset = run.optimization.indexsets.get(name=f"type_{name}")
         except IndexSet.NotFound:
-            category_indexset = self.index[ms].optimization.indexsets.create(
-                name=f"type_{name}"
+            category_indexset = IndexSet(
```
glatterf42, line 1138: Here I'm creating facade-level object from `_backend` functions so that subsequent usage of the items finds the property types it expects.

ixmp/tests/backend/test_ixmp4.py
```diff
@@ -260,13 +260,14 @@ def test_item_delete_elements(self, ixmp4_backend: Any, scenario: Scenario) -> N
...
         indexset_data = "foo"
-        indexset = run.optimization.indexsets.create("Indexset")
-        indexset.add(data=indexset_data)
-        table_data = {indexset.name: [indexset_data]}
-        table = run.optimization.tables.create(
-            "Table", constrained_to_indexsets=[indexset.name]
-        )
-        table.add(data=table_data)
+        table_data = {"Indexset": [indexset_data]}
+        with run.transact("Test item_delete_elements"):
```
glatterf42, line 264: With these tests, there's not much of a difference whether we use `run.transact()` or `_backend` functions and I found this quicker to add.

ixmp/tests/core/test_scenario.py
```diff
@@ -818,7 +818,8 @@ def test_filter_str(scen_empty: "Scenario") -> None:
 
     # Parameter defined over 's'
     p = pd.DataFrame.from_records(
-        zip(elements, [1.0, 2.0, 3.0]), columns=["s", "value"]
+        # zip(elements, [1.0, 2.0, 3.0]), columns=["s", "value"]
```
glatterf42, line 821: Another line I forgot to delete, sorry.

glatterf42 (commented):
Of course, bumping ixmp4 to >0.11.0 solves a few mypy issues, but introduces the requirement of `Run.lock` for some `.optimization` operations, extending the scope of this PR. 

However, this adjustment would soon be required, either way, so I did it here. The main thing we need to figure out/decide on is how we want to handle the `transact()` context. Within ixmp4, this is the only official way to lock and unlock a `Run`; the `Run._lock()` and `Run._unlock()` functions are marked as intended-private. 
The way I've used most of the time here is to directly call the `_backend` functions: `Run.lock` only exists in ixmp4's facade layer, the RestAPI and DB layer don't have that, allowing this circumvention.

glatterf42:
Thanks for the hints :)

1. I didn't know there are ISO dates, I'll adjust the title.
2. I'll also adjust the description :)
3. The codecov/project check passed before already, so I guess it wouldn't have been strictly necessary to address the codecov/patch check. The failure simply resulted from the fact that most lines changed in this PR belong to the few that were untested as of yet. But since we always strive for more test coverage, I wrote up a few test cases which should take care of most of these lines :)

ixmp/tests/backend/test_ixmp4.py
```diff
@@ -284,12 +285,69 @@ def test_item_delete_elements(self, ixmp4_backend: Any, scenario: Scenario) -> N
...
         run = ixmp4_backend.index[scenario]
-        indexset = run.optimization.indexsets.create("Indexset")
+        with run.transact("Test delete_item"):
+            indexset = run.optimization.indexsets.create("Indexset")
         ixmp4_backend.delete_item(s=scenario, type="set", name=indexset.name)
 
         # Test there are no 'sets' on scenario anymore
         assert ixmp4_backend.list_items(s=scenario, type="set") == []
 
+    def test_cat_set_elements(self, ixmp4_backend: Any, scenario: Scenario) -> None:
```
khaeru, line 295: Is there a particular reason this is a test only of IXMP4Backend?

I can imagine some possibilities, but can't tell which applies:

- The set-up for this test can only be performed on IXMP4Backend; JDBCBackend lacks features needed to do it.
  - IMO in such a case, it should be a common/parametrized test with an appropriate XFAIL mark for IXMP4Backend.
- The outcome/behaviour differs between JDBCBackend and IXMP4Backend.
  - This could also be a common test, with `is_ixmp4backend()` used to make the right assertion(s) in each case.

ixmp/backend/ixmp4.py
khaeru: I'm not sure I understand this. If ixmp4 "do[es]n't have a function like `remove_solution()`, how does the call in the next line succeed?

ixmp/backend/ixmp4.py
khaeru: That would be good as a comment, if you don't mind.

khaeru (approved):
Functionally and in terms of code quality this seems fine. I am mainly concerned that we adequately document the nuances of ixmp4 behaviour/data model so that it is clear what the IXMP4Backend code *should* do. That way we can look at what it *is* doing and satisfy ourselves that it's correct.

We should also consider that this will need to be maintained over time and updated to follow future changes in ixmp4, so the more clear we can make the intent, the better.

ixmp/backend/ixmp4.py
glatterf42: Following the explanation below, the above should make more sense: `remove_solution()` is a convenience function in the facade layer, which doesn't have a 1:1 equivalent in the backend layers. Instead, it is for now translated to two loops, removing the data of Variables and Equations one by one with appropriate backend calls.

As an aside, I'm aware that this is inefficient and I have some ideas of have to improve this. Most prominently, something like a `bulk_update()` setting `data = {}` should work, or maybe I should just write a new SQL query achieving the same without having to run through our data validation process etc.

ixmp/tests/backend/test_ixmp4.py
glatterf42: Only that I wanted to write a test covering all lines in the `IXMP4Backend.cat_set_elements()` function, but I can try how it works on JDBC :)

On your first possibility, I'm assuming you mean we would XFAIL the JDBCBackend case if it doesn't work.

ixmp/backend/ixmp4.py
meksor: Hi, that's fine but don't forget `.lock_with_timeout()`

ixmp/tests/backend/test_ixmp4.py
glatterf42: The actual case is: `ixmp_source` doesn't allow `cat_set_elements()` to be called for non-MESSAGE scenarios: https://github.com/iiasa/ixmp_source/blob/889b51f7731b3fdfed2e241c3d6596723e83202e/src/main/java/at/ac/iiasa/ixmp/objects/Scenario.java#L492

At the same time, the `JDBCBackend` function does not include any logic, it just forwards the information to the java method: https://github.com/iiasa/ixmp/blob/1310f2dc648a57d3ab7c07ba0b5e4ce055a4a408/ixmp/backend/jdbc.py#L1428

So a test on the JDBCBackend is not all that useful and performed by [`test_add_cat` in message_ix](https://github.com/iiasa/message_ix/blob/8f55fcf3779a6c5ee9884705485741580b69e8c0/message_ix/tests/test_core.py#L376).

The tests of backend functions are mostly in `test_jdbc`, so I can leave this test there and still mark it as `pytest.mark.ixmp4` with a note on why we're limiting this test case. Alternatively, I can maybe XFAIL or SKIPIF the test or use `is_ixmp4backend()` inside the test to either use `with pytest.raises()` or run the test as it works on ixmp4.

ixmp/backend/ixmp4.py
```diff
@@ -19,6 +19,12 @@
 from ixmp4.core.optimization.scalar import Scalar, ScalarRepository
 from ixmp4.core.optimization.table import Table
 from ixmp4.core.optimization.variable import Variable
+from ixmp4.data.abstract.optimization.indexset import (
```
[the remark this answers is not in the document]
khaeru: Okay, this helps. To the point of:

> Why and when does the IXMP4Backend code need to use one versus the other? 

I sort of sense that the facade layer(s) automate or mandate locking, in some way (one of the mentioned "conveniences") but the underlying layer (and classes of the same names) do not. So is the answer to the question roughly: "When the behaviour needs to be invoked without the automatic/mandatory locking, the underlying layer is used; otherwise the facade layer"—or am I still not getting it?

PR #596, opened 2025-09-09
Bump minimum required version of ixmp4

glatterf42:
The recent bump of allowed ixmp4 versions adds support for later versions, but in some aspects, these versions are incompatible with earlier versions. As @khaeru rightly notes, we don't need to support all versions of the (currently still unstable) ixmp4 API. So this PR bumps the minimum required version to save ourselves this unnecessary maintenance work.

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just handling dependencies.
- ~[ ] Update release notes.~ Just handling dependencies.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
     additional_dependencies:
     - genno
     - GitPython
-    - "ixmp4 >= 0.10, < 0.13"
+    - "ixmp4 >= 0.12, < 0.13"
     - nbclient
     - pandas-stubs
     - pytest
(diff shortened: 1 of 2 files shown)


khaeru (approved):
Thanks! Let's merge this once the tests pass.

At some point before the 3.12 release, we might edit/add to the migration roadmap as we discussed, e.g. to clarify that ixmp v3.11 ran with ixmp4 v0.10, and ixmp v3.12 will run with ixmp4 v0.12 (or perhaps a subsequent version, depending on how fast that goes in the next few months), and that users shouldn't expect stability in behaviour of the backend or its database schema.

I only mention this so we remember to do it; doesn't need to be in this PR.

Issue #75, opened 2018-07-30
Move `clone()` with kwarg `first_model_year` to `message_ix.Scenario`


danielhuppmann:
Calling the function `clone()` with a kwarg `first_model_year` on a non-MESSAGE-scheme Scenario will raise an error message, because it only applies there. Therefore, this option should be moved to the `message_ix.Scenario` class.

PR #601, opened 2025-10-14
Enhance functionality of `IXMP4Backend`

glatterf42:
This PR expands the functionality of `IXMP4Backend`, bringing it closer to full parity with `JDBCBackend`. It does so by:

- adding implementations for some functions
- clarifying that some functions won't be supported 
- migrating the ixmp4 tests to running on a PostgreSQL instance (enabling full functionality from ixmp4 in terms of versioning and rollback logic)

---

#### Setting up Postgres on GitHub Actions

The GHA runners only provide Postgres installations on Windows and Linux, but not on Macos. Thus, I'm trying to use [this third-party action](https://github.com/ikalnytskyi/action-setup-postgres), which works on all OSs and is actively maintained. 

## How to review

- Read the `#NOTE`, `#TODO` and `#FIXME` marks in the diff and respond if necessary.
- Check the pytest marks `@pytest.mark.ixmp4`, `@pytest.mark.ixmp4_not_yet`, and `@pytest.mark.ixmp4_never` to see if the usage/explanations are clear.
- Read the [documentation of differences between ixmp4 and JDBC](https://iiasa-energy-program-message-ix--601.com.readthedocs.build/projects/ixmp/en/601/api-backend.html#differences-between-jdbcbackend-and-ixmp4backend) and see if it's clear or if anything might be missing.
- Respond to comments I made on the files here on GitHub.

TBD:

- Once https://github.com/iiasa/message_ix/pull/981 is merged and we can link to the new docs section: Try setting up a local ixmp4 postgres DB and running some tests on that

## PR checklist

- [x] Add or expand tests ~; coverage checks both ✅~ (see [below](https://github.com/iiasa/ixmp/pull/601#issuecomment-3522029169))
- [x] Update release notes.

### DIFF
--- a/ixmp/testing/__init__.py
+++ b/ixmp/testing/__init__.py
@@ -74,6 +74,7 @@ from .resource import resource_limit
 if TYPE_CHECKING:
     from _pytest.mark import ParameterSet
     from ixmp4.core import Run
+    from sqlalchemy import Engine  # noqa: F401
 
     try:
         # Pint 0.25.1 or later
@@ -107,6 +108,10 @@ __all__ = [
 
 #: :any:`True` if testing is occurring on GitHub Actions runners/machines.
 GHA = "GITHUB_ACTIONS" in os.environ
+
+# Pytest stash keys
+KEY_ENGINE = pytest.StashKey["Engine"]()
+KEY_IXMP4_PG_NAME = pytest.StashKey[str]()
 
 _uname = platform.uname()
 
@@ -179,7 +184,7 @@ def pytest_sessionstart(session: pytest.Session) -> None:
     - :data:`.jdbc._GC_AGGRESSIVE` is set to :any:`False` to disable aggressive garbage
       collection, which can be slow.
     """
-    from ixmp.backend import available, jdbc
+    from ixmp.backend import jdbc
 
     if not session.config.option.ixmp_user_config:
         ixmp_config.clear()
@@ -194,34 +199,29 @@ def pytest_sessionstart(session: pytest.Session) -> None:
         from xdist import get_xdist_worker_id
 
         db_name = f"ixmp_test_{get_xdist_worker_id(session)}"
-
-        engine = create_engine(
-            "postgresql+psycopg://postgres:postgres@localhost:5432/postgres",
-            isolation_level="AUTOCOMMIT",
-        )
+        url = "postgresql+psycopg://postgres:postgres@localhost:5432/postgres"
+        engine = create_engine(url, isolation_level="AUTOCOMMIT")
         with engine.connect() as connection:
             connection.execute(text(f"CREATE DATABASE {db_name}"))
 
-        engine.dispose()
+        # Store for pytest_sessionfinish()
+        session.config.stash[KEY_ENGINE] = engine
+        session.config.stash[KEY_IXMP4_PG_NAME] = db_name
 
 
 def pytest_sessionfinish(
     session: pytest.Session, exitstatus: int | pytest.ExitCode
 ) -> None:
-    if "ixmp4" in available():
-        from sqlalchemy import create_engine, text
-        from xdist import get_xdist_worker_id
+    try:
+        from sqlalchemy import text
 
-        db_name = f"ixmp_test_{get_xdist_worker_id(session)}"
+        engine = session.config.stash[KEY_ENGINE]
+        db_name = session.config.stash[KEY_IXMP4_PG_NAME]
 
-        engine = create_engine(
-            "postgresql+psycopg://postgres:postgres@localhost:5432/postgres",
-            isolation_level="AUTOCOMMIT",
-        )
         with engine.connect() as connection:
             connection.execute(text(f"DROP DATABASE {db_name}"))
-
-        engine.dispose()
+    except (ImportError, KeyError):
+        pass
 
 
 def pytest_report_header(config: pytest.Config, start_path: Path) -> str:


ixmp/tests/test_integration.py
```diff
@@ -66,25 +66,26 @@ def test_run_clone(
...
-    assert np.isnan(scen4.var("z")["lvl"])
-    assert_frame_equal(
-        scen4.timeseries(iamc=True),
-        TS_DF_CLEARED.assign(scenario=pd.Series([scen.scenario, scen.scenario])),
-    )
-
-
-# FIXME Fix IXMP4Backend return value for s.var()["lvl"] so that np.isnan() accepts it
-@pytest.mark.jdbc
+    # NOTE This works on JDBC, but is part of a stack violation; thus, ixmp4 is not
```
glatterf42: I'm not adding a documentation entry about this since we aim to address https://github.com/iiasa/ixmp/issues/75 before releasing the documentation of differences between ixmp4 and JDBC.

glatterf42:
For the coverage, the overall project is fine, so the fact that it seems lacking is simply because the parts of the code that this PR touches are not covered as well as the rest.
Additionally, I went over the lines flagged as not covered, and it seems to me that they are fine; they are only safeguard lines that I thought would be necessary for the test suite to pass, but if they are not called, that assessment seems wrong. We could remove them or keep them, of which I'd prefer the latter.

doc/api-backend.rst
```diff
@@ -114,6 +115,164 @@ IXMP4Backend
...
+^^^^^^^
+
+On JDBCBackend, each region can have a "parent" and a "hierarchy"
+and may have a "synonym" defined for itself.
+Notably, when a region is part of a "hierarchy", a "parent" must be defined.
+
+On ixmp4, each region must be part of a "hierarchy", 
+but does not have either "parent" nor "synonym".
+
+For backward compatibility, IXMP4Backend accepts the same parameters for region-related function as JDBCBackend.
```
khaeru: In general, it will be better to write this from the perspective of a user of ixmp or someone who is using ixmp through message_ix. They may not know/care about the background classes, but will need to be aware that `Platform.add_region()` is affected.

This will also help them search and find this docs section if they try to use that particular method and something unexpected happens.

ixmp/tests/test_cli.py
```diff
@@ -34,8 +34,8 @@ def test_versiontype() -> None:
         vt.convert("xx", None, None)
 
 
-# FIXME This should work for IXMP4Backend, I think
-@pytest.mark.jdbc
+# FIXME The URL needs to be different for ixmp4; can you target model/scenario on pg?
```
khaeru: The ``ixmp://`` URL scheme uses the platform name appearing in the user's ixmp/config.json as the first part ('host'). So in the case of the local test platform, this should be `ixmp4-local`.

ixmp/tests/report/test_operator.py
```diff
@@ -236,18 +236,26 @@ def test_store_ts(caplog: pytest.LogCaptureFixture, test_mp: "Platform") -> None
 
     # A message is logged
     r = caplog.record_tuples[-1]
+    error = (
```
khaeru: This is a case where the shim *could* catch and re-raise the exception in order to satisfy the test expectation, instead of adjusting the test. It's still passable to do the latter.

I hope there is little or no downstream code that relies on the specific class and text of this exception; if there is, it can be adjusted following this example.

ixmp/backend/base.py
```diff
@@ -1338,6 +1342,29 @@ def cache_get(
         else:
             raise KeyError(ts, ix_type, name, filters)
 
+    def maybe_get_cache(
+        self, ts: TimeSeries, ix_type: str, name: str, filters: "Filters"
+    ) -> "SetData | ParData | SolutionData | None":
+        try:
```
khaeru: This method should have a docstring.

doc/api-backend.rst
khaeru: I'll join in on the PR to copyedit this docs section a bit.

ixmp/backend/jdbc.py
```diff
@@ -1165,32 +1165,20 @@ def item_get_elements(
...
-    # FIXME reduce complexity 18 → ≤13
+    # FIXME reduce complexity 16 → ≤13
     def item_get_elements(  # noqa: C901
         self, s: Scenario, ix_type: str, name: str, filters: "Filters" = None
     ) -> "SetData | ParData | SolutionData":
         if filters:
             # Convert filter elements to strings
             filters = {dim: as_str_list(ele) for dim, ele in filters.items()}
 
-        try:
```
khaeru: Good call to move this, it was in the wrong place.

ixmp/core/timeseries.py
```diff
@@ -103,6 +104,9 @@ def __init__(
             # annotation will be "" if None is provided, convince type checker
             assert annotation is not None
             self.platform._backend.init(self, annotation)
+
+            if is_ixmp4backend(self.platform._backend):
+                self.platform._backend.index[self]._lock()
```
khaeru: Can this also go in IXMP4Backend.init()?

ixmp/testing/__init__.py
```diff
@@ -183,6 +188,40 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
+def pytest_sessionfinish(
+    session: pytest.Session, exitstatus: int | pytest.ExitCode
+) -> None:
+    if "ixmp4" in available():
+        from sqlalchemy import create_engine, text
+        from xdist import get_xdist_worker_id
+
+        db_name = f"ixmp_test_{get_xdist_worker_id(session)}"
+
+        engine = create_engine(
```
khaeru: Instead of having very similar code to above, the [`pytestconfig.stash`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.Config.stash) feature can be used.

ixmp/testing/__init__.py
```diff
@@ -209,13 +248,25 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
             elif "jdbc" in marker_names:
-                # This marker means "not (yet) implemented/supported on IXMP4"
-                i = argvalues.index("ixmp4")
-                argvalues[i] = pytest.param("ixmp4", marks=MARK["IXMP4Backend NI"])
+                # Same as "ixmp4" marker, but for JDBCBackend
+                argvalues.remove("ixmp4")
 
-            if "ixmp4" in argvalues and "ixmp4_209" in marker_names:
-                i = argvalues.index("ixmp4")
+        if "ixmp4" in argvalues:
```
khaeru: I suspect this can be combined with the previous if-block; will give it a try.

ixmp/testing/__init__.py
```diff
@@ -579,6 +640,7 @@ def _platform_fixture(
     tmp_env: os._Environ[str],
     test_data_path: Path,
     backend: Literal["jdbc", "ixmp4"],
+    worker_id: str,
```
khaeru: Since this is universally available with xdist, I don't think it needs to be passed in as an argument.

ixmp/tests/core/test_platform.py
```diff
@@ -22,44 +22,6 @@
     from ixmp.types import PlatformInitKwargs
 
 
-@pytest.mark.usefixtures("tmp_env")
```
khaeru: I'm realizing I don't know where I picked this up from, but the general order of contents in a Python module is:

- Globals
- Classes
- Functions

Unless there's a reason to move this class to the end of the file, let's keep it where it is. That would also make it easier to see if there are changes to particular lines.

khaeru (changes requested):
Thanks for this PR!

Due to the size, I was not able to go linewise over all the additions, particular to `ixmp.backend.ixmp4`. Where those were entirely new methods, I guessed they were implementations of base.Backend methods that did not previously exist, and in some sense "the proof is in the pudding [tests]"—so I focused on those.

There are a number of comments in-line; in some places I mention that it will be more efficient if I step in and make the changes directly (e.g. in the documentation) rather than write in abstract about what should change. I will only be able to start on that in a few hours, after dealing with some other pressing tasks.

ixmp/testing/__init__.py
khaeru: > So I would also need to store this key somehow for later usage.

Yes, that's the intended usage. Maybe https://docs.pytest.org/en/stable/example/simple.html#making-test-result-information-available-in-fixtures is a more direct example in the pytest docs.

See also https://github.com/khaeru/sdmx/blob/abf1d4301decb645cfe4bfac926c5a4f18150470/sdmx/testing/__init__.py#L33-L36

PR #602, opened 2025-10-14
Confirm support for Python 3.14 and drop 3.9

khaeru:
Python 3.14 [was released 2024-10-07](https://peps.python.org/pep-0745/). This PR is to:

- [x] Check/confirm that ixmp is compatible with the new version by updating CI workflows.
  - [x] Update the GitHub ruleset config to require 3.14 jobs to pass, instead of 3.13.
- [x] Check/confirm the existence of compatible releases of upstream dependencies. If those releases are not immediately available, the PR can aid to monitor and track upstream issues.
  - khaeru/genno#179. Commits in that PR can be used as an example/guide to address any tasks also relevant for this PR.
  - iiasa/ixmp4#209 and unionai-oss/pandera#2158. These are addressed by XFAIL marks on affected tests, which can be later removed when the upstream issues are resolved.
- Drop support for Python 3.9, which reached end-of-life (no more updates, even security updates) [on 2025-10-07](https://peps.python.org/pep-0569/#lifespan).
  - [x] Remove any code marked as for Python 3.9 compatibility only. Depending on these changes, part or all of the functionality may remain usable with Python 3.9 if users have extreme need.
  - [x] Make any general clean-ups or improvements possible with [features available from Python 3.10](https://docs.python.org/3/whatsnew/3.10.html).
    - [x] Use the type union operator instead of typing.Optional and typing.Union. This is achieved by running `ruff check --select F401,UP007,UP045 --fix` or similar multiple times, plus some small manual clean-ups. See [the ruff docs](https://docs.astral.sh/ruff/rules/#pyupgrade-up).
    - [x] Use `match`/`case` instead of some `if` statements with multiple `elif` blocks.

Analogous to #544.

## How to review

- Read the diff.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -33,6 +33,8 @@ Migration notes
 All changes
 -----------
 
+- :mod:`ixmp` is tested and compatible with `Python 3.14 <https://www.python.org/downloads/release/python-3140/>`__ (:pull:`602`).
+- Support for Python 3.9 is dropped (:pull:`602`), as it has reached end-of-life.
 - Improve :class:`.IXMP4Backend` (:pull:`581`):
 
   - Support creation and modification of 0-dimensional parameters (:class:`ixmp4.Scalar`).


khaeru:
With the other changes in the PR, one test appears flaky, as it fails on a mix of OSes and Python versions that don't have any clear pattern:
```
FAILED ixmp/tests/core/test_scenario.py::TestScenario::test_add_par2[ixmp4] - IndexError: list index out of range
```

I'll mark this as such for now.

ixmp/core/platform.py
```diff
@@ -162,8 +155,8 @@ def get_log_level(self) -> str:
     def scenario_list(
         self,
         default: bool = True,
-        model: Optional[str] = None,
-        scen: Optional[str] = None,
+        model: str | None = None,
+        scen: str | None = None,
```
glatterf42:
```suggestion
        scen: str | None = None
```
I'm wondering if these annotations are now short enough to fit into a single line. If you remove the comma as in the suggestion, does ruff change the formatting or just re-introduce the comma?

ixmp/tests/backend/test_ixmp4_io.py
glatterf42: Maybe we could also do an `if TYPE_CHECKING` import of `ixmp4.core.Run` and annotate these fixtures properly :)

ixmp/testing/__init__.py
```diff
@@ -69,6 +69,9 @@
 from .jupyter import get_cell_output, run_notebook
 from .resource import resource_limit
 
+if TYPE_CHECKING:
+    from ixmp4.core import Run
```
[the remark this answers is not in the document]
khaeru: In general:
- We assume that people doing type checking will have all the type checking dependencies specified in the .pre-commit-config.yaml file. This is a full/long list of dependencies, similar in scope to `ixmp[tests]`.
- `.testing` should be importable (but perhaps not fully usable) even if the user has only the minimum set of `ixmp` dependencies.
- For the test suite *per se* (`.tests` and submodules), probably yes, we assume the user has `ixmp[tests]` installed.

I am of mixed minds on using `if TYPE_CHECKING:` in the test suite. On the one hand, per the above logic, it's not necessary to guard against packages not installed at (test) run time. But on the other, it helps to clarify which imports are _used_ in a module vs. only there for type checking. Keeping the former list smaller helps in thinking about code.

PR #610, opened 2025-11-14
Enable cloning from jdbc to ixmp4

glatterf42:
In order to migrate from Oracle to PostgreSQL, we'll need a way to migrate (at least the most important) Scenarios. This PR adds a function that achieves just that, though there's still room for some improvement plus one open question.

@volker-krey In theory, you could use this branch to try cloning something from "ixmp_dev" to a local ixmp4/PostgreSQL DB (see https://github.com/iiasa/message_ix/pull/981 on how to set one up).

More info:
- The branch depends on changes in the branch for iiasa/ixmp4#208, and can only be run with that version of ixmp4—not `main`, and not v0.14.0 from PyPI.
- Supersedes/replaces #424.

## Important question

- (Deferred to a subsequent PR.) Users might want to drop the timeseries/IAMC data that's considered "solution", but at least in the Python API, there's no way to read only timeseries data with `meta=True`. Should I try to hack something together using ixmp_source directly or are we okay with saying one can only clone full Scenarios from JDBC to ixmp4 (in that regard; dropping the optimization solution is easy)?
  - @khaeru: “It will be straightforward to provider users with utilities that can selectively remove time series data in scenarios that have been cloned to ixmp4 databases. That is, it can be a 2-step operation. Since the cross-backend clone will only happen in the context of careful migration at first, I think this will be fine.” 

## Possible improvements

- (Deferred.) Use `bulk_insert()` functions on the ixmp4 side: this is much more involved and requires implementing such functions in ixmp4 first, so if we don't mind performance so much, we could start an experiment with the current implementation.
  - @khaeru: “Given we are still waiting on merge of iiasa/ixmp4#208, let's not delay this PR for work that hasn't started yet upstream.”
- [x] Add a docstring. Not sure this is needed, though, as the function should be self-explanatory (if long) and is marked as private.

## How to review

- Read the diff and note that the CI checks (including the new test!) all pass.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -41,7 +41,9 @@ All changes
 - Improve :class:`.IXMP4Backend` (:pull:`581`):
 
   - Support creation and modification of 0-dimensional parameters (:class:`ixmp4.Scalar`).
-  - Streamline and fix various functions to enable more tests to pass (:pull:`601`, :pull:`611`).
+  - Streamline and fix various functions to expand coverage of the :mod:`ixmp` API (:pull:`601`, :pull:`611`).
+    All remaining differences between :class:`.JDBCBackend` and :class:`.IXMP4Backend` are documented.
+  - :class:`.Scenario.clone` can clone a Scenario *from* JDBCBackend *to* IXMP4Backend (:pull:`610`).
 
 - New method :meth:`.Scenario.iter_par_data` (:pull:`581`).
   :meth:`.Scenario.items` no longer supports iterating over item *contents*.
@@ -64,7 +66,6 @@ All changes
   Calling this method emits :class:`DeprecationWarning`,
   and the method will be removed in a future version of :mod:`ixmp`.
 - Improve performance of :meth:`.Scenario.remove_par` and :meth:`.Scenario.remove_set` (:pull:`598`).
-- Document all differences between :class:`.JDBCBackend` and :class:`.IXMP4Backend` (:pull:`601`).
 
 .. _v3.11.1:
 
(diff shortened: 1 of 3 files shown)


khaeru:
Thanks for the solid start here @glatterf42, especially for the test setup to confirm that the operation works.

I can join in/take over to make some adjustments. At a glance, I have these in mind:
- To align with existing separation-of-concerns, TimeSeries.clone() should handle clone of time-series data associated with the TimeSeries class, and Scenario.clone() should handle clone of model data. Each can make necessary adjustments on the target Platform for consistency.
- I think the proper location for clone code varies according to the situation:
  - Cloning within 1 Backend or across 2 Backends of the same type: should be in the respective backend module and make maximal use of low-level code for performance.
  - Cloning across 2 Backends of different type: should be high-level code using only the Platform, TimeSeries, and Scenario APIs, i.e. each Backend should not (need to) be aware that there is a distinct Backend class on the other end of the clone.

I see see the new high-level method Scenario.iter_item_data() is called from within a _lower-level_/underlying class, JDBCBackend. The latter adjustment would avoid having this somewhat backward relationship.

With the tests as a guardrail, I will try to make these changes as soon as I have a chance.

khaeru (approved):
Thanks @glatterf42 for developing this key feature and especially the tests for it, which made it easy to do some reorganization. This paves the way for similarly testing message_ix.Scenario.clone(): first to test databases, and eventually between centrally-operated IIASA Oracle and PostgreSQL databases.

At this point, it seems like ixmp *per se* is in good shape, and we are more waiting on:
1. Review and merge of iiasa/ixmp4#208 by maintainers.
2. Discussion on setting up and administering those central databases.
3. PRs for message_ix and message-ix-models.

…none of which should be in scope for the next week. So I will merge this PR, and then we can revise if necessary as those other pieces catch up. In particular, if the upstream PR is still pending by the time the 3.12 milestone comes (now scheduled 23 January), I can put in appropriate safeguard or warnings for the next release

